PR #31, opened 2014-12-05
BUGFIX: Extensible Match Filter based on DN attributes not working.

cwaldbieser:
I noticed there was a bug in the extensibleMatch filtering (ref https://tools.ietf.org/html/rfc4511#section-4.5.1).  I added the code to correct and a couple of unit tests.

Thanks,
Carl


psi29a:
@cwaldbieser can you fix this?

cwaldbieser:
Yes.  It makes sense.  I wouldn't want to catch a Ctrl-C SIGINT.  Should be
a simple correction.  I should be able to get to it tomorrow.

PR #34, opened 2015-01-08
Don't leave LDAP clients hanging if proxied server cannot be contacted

cwaldbieser:
I noticed that if the proxied LDAP server cannot be contacted, the client ends up hanging.  I made corrections to `proxybase.py` to correct this and added some tests.  pyflakes was happy.  I ran pep8 on the test code, and it had a long list of complaints, which I corrected except for line lengths.  This did result in some line breaks around classes/functions I am not sure fit in with the current code base.  If those are problematic, let me know.

Thanks,
Carl


psi29a:
Good find!

PR #35, opened 2015-01-27
Server pwdchange

cwaldbieser:
I noticed that issuing an LDAP password change request would seem to work for an LDAP server using an ldiftree backend (`ldaptor/ldiftree.py`).  However, if you subsequently issued an LDAP search for the entry which just had its password changed, the `userPassword` attribute was unchanged.

I determined this was because `LDAPServer` was not calling `commit()` like it was doing for an LDAP modify operation.

The server tests use the in-memory back end.  `commit()` is a no-op for that back end, so I had to wrap the commit method of the entry being tested to verify that commit was being called.

It was not clear what the Deferred that is returned from `commit()` was supposed to fire-- the in-memory back end returned the entry itself, while the ldiftree didn't return anything.  There was a curious comment that suggested that commit should indicate whether it succeeded or failed, so I chose to have it fire a simple boolean value and patched that up.

I made lots of formatting changes on the files I touched to keep `pyflakes` and `pep8` happy.

### DIFF
--- a/ldaptor/entry.py
+++ b/ldaptor/entry.py
@@ -1,9 +1,10 @@
-import random, base64
-from zope.interface import implements
-from twisted.internet import defer
-from twisted.python.util import InsensitiveDict
+import base64
+import random
 from ldaptor import interfaces, attributeset, delta
 from ldaptor.protocols.ldap import distinguishedname, ldif, ldaperrors
+from twisted.internet import defer
+from twisted.python.util import InsensitiveDict
+from zope.interface import implements
 
 try:
     from hashlib import sha1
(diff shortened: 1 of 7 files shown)


cwaldbieser:
One more thing I forgot to mention-- I noticed that ldiftree only implemented salted, SHA1 hashes.  It used a scheme similar to OpenLDAP which supports a number of different password storage formats.  I added the plain text format because it is convenient to work with when writing tests or developing.  It would be nice to support pluggable password storage formats for a simple BIND in the future.

psi29a:
I'll review this in a bit, but can you create an issue for pluggable password storage format, so we can track it.

Thank you again for the contributions! :)

psi29a:
@coveralls would you be so kind as to add yourself to the docs/source/AUTHORS.rst file. :)

psi29a:
@cwaldbieser would you be so kind as to add yourself to the docs/source/AUTHORS.rst file. :)

Issue #38, opened 2015-05-05
ldapsearch against ldaptor always returns all attributes


the2nd:
Hi,
if i do an ldapsearch agains ldaptor which requests just one (or some) attributes ldaptor seems to always return the complete ldif of the found object. Using the same seach against openldap only the "dn" and the requested attributes are returned.

You should be able to reproduce with the following steps:

Try ldapsearch which requests only specific attribute(s) with ldaptor:
1. run ldaptest.py
2. on second console run ldapsearch command that requests only "uid" attribute: ldapsearch -H ldap://localhost:8080 -b "dc=de" -w afdg -x uid
3. the ldif returned from ldapsearch contains all object attributes!

Try ldapsearch which requests only specific attribute(s) with openldap:
1. copy slapd.conf to /etc/openldap (or whatever directory your distribution uses)
2. import ldap tree from ldaptor: ldapsearch -H ldap://localhost:8080 -b "dc=de" -w afdg -x | slapadd -l /dev/stdin
3. start slapd (e.g. /etc/init.d/slapd start)
4. run ldapsearch command that requests only "uid" attribute: ldapsearch -H ldap://localhost -b "dc=de" -D 'cn=manager,dc=example,dc=de' -w root -x uid
5. the ldif returned from ldapsearch contains only "dn" and "uid" attributes

ldaptor.py: https://bpaste.net/show/3f5174d8ec15
slapd.conf: https://bpaste.net/show/c334ec415850

reference: http://twistedmatrix.com/pipermail/twisted-python/2015-May/029447.html

psi29a:
@the2nd Does https://github.com/twisted/ldaptor/pull/40 fix your problem? :)

the2nd:
Yes #40 fixes the issue. Thanks a lot! :)

PR #40, opened 2015-05-12
BUGFIX: LDAP search should return only the attributes requested.

cwaldbieser:
Should return all attributes if no attributes are requested.

This should correct issue #38.
I made the correction and wrote a couple tests.

Thanks,
Carl


psi29a:
Great work! Thanks again.

PR #41, opened 2015-05-29
Added an LDAP Server, which merges requests

cbartz:
Think of it as a proxy, which forwards requests to multiple LDAP servers, and merges the responses.
Only Bind and Search requests are supported.

Source Code is based on "proxy.py" and "proxybased.py".


psi29a:
Looks good! Thank you for the PR! :)

PR #42, opened 2015-07-02
Cookbook

cwaldbieser:
I tried updating the documentation to be a little more straightforward.  I thought having both quickstart client and server sections would be useful for folks that want to dive right in.  I thought the introductory text could be better organized.  I added a "Cookbook" section for putting various short Ldaptor examples.  I was inspired by the O'Reilly Python Cookbook book, which I thought was a great way to learn how do accomplish various tasks with Python by example.

The documentation is by no means complete-- I am thinking of it as more of a starting point.

I apparently also snuck in a correction to errors I had made in the `proxybase` test code docstring.

### DIFF
--- a/docs/source/addressbook-example.rst
+++ b/docs/source/addressbook-example.rst
@@ -1,9 +1,16 @@
+==================================
 Creating a simple LDAP application
 ==================================
 
-LDAP presents a distributed tree of information
------------------------------------------------
+--------------
+An LDAP Primer
+--------------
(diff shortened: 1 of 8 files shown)


cwaldbieser:
Looks like travis failed because a virtualenv didn't download.  Not sure there is anything I can do about that.

psi29a:
Restarting that particular build, but I think we shouldn't be held up by doc fixes. :)

Thank you for taking the time to do this!

PR #43, opened 2015-07-06
Proxybase - use `clientConnector` for making client connections, updated some docs

cwaldbieser:
I realized that the `ProxyBase` class assumed a particular kind of connection to the proxied server using the `ldaptor.protocols.ldap.ldapconnector.LDAPClientCreator` or something very similar.  This made it difficult to integrate it with connecting using Twisted client endpoints.  Since there has not been a release since `ProxyBase` was added, I broke the original API and replaced the relevant properties with a `clientConnector` property that is a callable that should take care of establishing the connection and returns a deferred that fires with a client protocol instance.

Related, I updated some of the documentation that referenced the old API.  

I also updated some of the documentation for clients and servers in general to use Twisted endpoints instead of `LDAPClientCreator`.  There were a couple reasons.  

Typical LDAP clients I have worked with just want to connect to a host and a port and maybe use LDAPS or STARTTLS.  They do not use a much more complex configuration than that, nor do they try to use DNS SRV records, which `LDAPClientCreator` throws into the mix.  In my opinion, this muddies the waters a bit for someone just trying to get started with ldaptor, and I think the documentation should reflect that.

On a technical level, `LDAPClientCreator` is based on the older Twisted `ClientCreator` which is somewhat less flexible than the newer client endpoints.  The `LDAPClientCreator` implementation also makes use of what appears to be a "private" class, `protocol._InstanceFactory`. 

This makes me think that unless someone *really* needs to use SRV lookups, the documentation should just steer them to making a more straightforward connection to the server.

### DIFF
--- a/docs/source/addressbook-example.rst
+++ b/docs/source/addressbook-example.rst
@@ -39,9 +39,9 @@ Writing things down, John Smith LDIF::
      ple lines as long as the non-first lines are inden
      ted in the LDIF.
 
-------------------------------------
+-------------------------------------
 Asynchronous LDAP Clients and Servers
-------------------------------------
+-------------------------------------
 
 Ldaptor is a set of pure-Python LDAP client and server protocols and libraries..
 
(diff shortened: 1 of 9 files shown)


psi29a:
Looks like it is broken on Twisted >= 13.0

psi29a:
pyflakes is also not happy ;)

pyflakes runtests: PYTHONHASHSEED='471523457'
pyflakes runtests: commands[0] | pyflakes ldaptor
ldaptor/test/test_proxybase.py:7: 'deferLater' imported but unused
ldaptor/test/test_proxybase.py:10: 'config' imported but unused
ldaptor/test/test_proxybase.py:72: undefined name 'WontConnectClientCreator'
ERROR: InvocationError: '/home/travis/build/twisted/ldaptor/.tox/pyflakes/bin/pyflakes ldaptor'pyflakes runtests: PYTHONHASHSEED='471523457'
pyflakes runtests: commands[0] | pyflakes ldaptor
ldaptor/test/test_proxybase.py:7: 'deferLater' imported but unused
ldaptor/test/test_proxybase.py:10: 'config' imported but unused
ldaptor/test/test_proxybase.py:72: undefined name 'WontConnectClientCreator'
ERROR: InvocationError: '/home/travis/build/twisted/ldaptor/.tox/pyflakes/bin/pyflakes ldaptor'

cwaldbieser:
Sorry about missing `pyflakes` on the test-- that should be corrected.

I think you meant Twisted <= 13.0.  I took a look at that, and it seems that `endpoints.connectProtocol()` was not introduced until Twisted 13.1.  It was a very short function, so I tried the import and if it failed, I just defined the function in the `ldapconnector` module.

I am not sure if there is a more typical way of handling this.  If so, let me know.

Thanks,
Carl

psi29a:
Yeah, sorry... was too early. ;)

Looks good otherwise!

PR #44, opened 2015-07-08
Documentation updates

cwaldbieser:
Primarily, I updated the "Creating a Simple Application ..." section of the docs.  The interactive examples were broken because they relied on a deprecated function in `twisted.trial` to get the results of deferreds at the interactive prompt.  Luckilly, twisted provides its own interactive shell *with* a reactor in the `twisted.conch` package, so I was able to salvage the interactive examples by adding a note to try them in that shell.

I also reorganized the sub-sections until I felt they were in the correct order.  

I changed the Sphinx config setting 'default' to 'alabaseter' because Sphinx was complaining that 'default' is going away and I should choose either 'alabaster' or 'classic'.  I am not partial to either-- just wanted to silence the warning.


psi29a:
Thank you very much for taking the time to do this!  :)  

It looks good to me.

PR #45, opened 2015-10-10
Complete the example with a missing module.

cwaldbieser:
The example was incomplete-- it was missing the example directory data.  I updated the docs to include the directory data.  Just a small update to the docs.

### DIFF
--- a/docs/source/cookbook/servers.rst
+++ b/docs/source/cookbook/servers.rst
@@ -15,6 +15,75 @@ A minimal LDAP DIT that stores entries in the local file system
 Code
 ''''
 
+First, a module that defines our DIT entries-- :file:`schema.py`
+
+.. code-block:: python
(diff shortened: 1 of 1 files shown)


cwaldbieser:
Not sure why the CI checks failed ...

psi29a:
Looks like we hit a wall on a few things,  the very latest version of Twisted (trunk) dropped Python 2.6 support. We'll have to adjust Tox to either not test this or put it under allowable failures.

PyPy is a mess right now because Cryptography requires a specific version that isn't available by default in Travis. This also means it needs to be moved to allowable failures in Tox, for now.

psi29a:
Merging this as it doesn't touch code.  I will go through ldaptor eventually and clean up tests now that py2.6 has been dropped on all distros.

PR #49, opened 2016-02-02
Import errors for UsageError in some scripts in bin

schlenk:
The ldaptor.usage module does not provide the UsageError exception. It is only in the twisted.python.usage module.

### DIFF
--- a/ldaptor/usage.py
+++ b/ldaptor/usage.py
@@ -1,6 +1,18 @@
 from twisted.python import usage, reflect
+from twisted.python.usage import UsageError
 from ldaptor.protocols import pureldap
 from ldaptor.protocols.ldap import distinguishedname
+
+__all__ = [
+    "Options",
+    "Options_base",
+    "Options_base_optional",
+    "Options_bind",
+    "Options_bind_mandatory",
+    "Options_scope",
+    "Options_service_location",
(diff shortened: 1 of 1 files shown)


cwaldbieser:
I am not sure I follow.  It looks like you took `usage.UsageError` in several files, imported the name directly, and replaced the occurrences with `UsageError`.  Is this just a style thing?

schlenk:
Usage isn't defined in ldaptor.usage, it is only defined in twisted.python.usage (which is used by ldaptor.usage internally).

    >>> import ldaptor.usage
    >>> ldaptor.usage.UsageError
     File "<stdin>", line 1, in <module>
     AttributeError: 'module' object has no attribute 'UsageError'
     >>> import twisted.python.usage
     >>> twisted.python.usage.UsageError
    <class 'twisted.python.usage.UsageError'>

psi29a:
I'm OK with this, if you can do the alternative fix. Can you do that instead please? 

Pay no attention to Travis at the moment, it is lying because py26 doesn't work with latest

schlenk:
Ok. I'll do the alternate fix (much simpler and more sane for sure.).

psi29a:
Thanks! :)

PR #52, opened 2016-04-14
Requested attributes fix and subtree build change

GrayAn:
1. Some clients request for all attributes by sending ['*'] as a list of required attributes (like Apache Directory Studio).
2. Recursive callback of _processOneChild in subtree method reaches recursion limit while searching through a long list of children. With recursion limit of 1000 it processes ~180 child entries before raising exception. I found no way to fix it without removing deferred getting of every next child.

### DIFF
--- a/ldaptor/entryhelpers.py
+++ b/ldaptor/entryhelpers.py
@@ -130,7 +130,7 @@ class SubtreeFromChildrenMixin(object):
             callback(self)
             d = self.children()
             def _gotChildren(children, callback):
-                for c in children:
+                for c in reversed(children):
                     c.subtree(callback)
             d.addCallback(_gotChildren, callback)
             return d
--- a/ldaptor/entryhelpers.py
+++ b/ldaptor/entryhelpers.py
@@ -129,9 +129,15 @@ class SubtreeFromChildrenMixin(object):
         else:
             callback(self)
             d = self.children()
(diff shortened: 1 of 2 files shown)


cwaldbieser:
* Specifying "*" in the attribute list is explicitly supported in RFC4511, so that seems like a good change.

* The change you made to correct the recursion limit-- something doesn't seem right to me.  The `subtree()` method returns a Deferred.  This suggests to me that the function might block.  Your change calls the children `subtree()` methods iteratively and *synchronously*.  That means if it does block, the whole server will be blocking.

I don't think you ran into this because the 2 existing DIT models, `inmemory.py` and `ldiftree.py`, don't block when they call their `children()` method.  However, I could imagine a DIT model based on a database backend that *could* block to fetch children.

I do think the existing code needs to be corrected.  If `children()` were asynchronous, it wouldn't make sense to wait for the result of each child before moving on to the next.  Maybe something like a DeferredList would be appropriate?

I'd separate this into to PRs, and give the second request some more consideration.

GrayAn:
- Moved changes in `subtree()` method to another branch and reversed the changes in "master" until we found the best solution. I'll experiment on asynchronous children iteration.
- Kept requested attributes fix.

psi29a:
Thanks. :)

PR #53, opened 2016-06-03
When parsing LDAPControls, criticality may not exist while controlValue still does.

aren:
See updated unit test.


psi29a:
LGTM :+1: 

Note to self, drop py26 support for latest twisted. ;)

psi29a:
Thank you!

PR #65, opened 2016-12-02
Fix extended response handling

im-0:
I tried to implement simplest LDAP proxy based on [example from documentation](http://ldaptor.readthedocs.io/en/latest/cookbook/ldap-proxy.html#code) and found that TLS client (`use_tls = True`) is not working with my server. Here is the output of example code:
```
2016-12-02 05:09:03+0300 [-] Log opened.
2016-12-02 05:09:03+0300 [-] ServerFactory starting on 10389
2016-12-02 05:09:03+0300 [-] Starting factory <twisted.internet.protocol.ServerFactory instance at 0x109102dd0>
2016-12-02 05:09:09+0300 [twisted.internet.protocol.ServerFactory] Starting factory <ldaptor.protocols.ldap.ldapconnector.OneShotFactory instance at 0x109125bd8>
2016-12-02 05:09:10+0300 [LDAPClient,client] Unhandled Error
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/twisted/python/log.py", line 84, in callWithLogger
       	    return callWithContext({"system": lp}, func, *args, **kw)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/twisted/python/log.py", line 69, in callWithContext
       	    return context.call({ILogContext: newCtx}, func, *args, **kw)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
       	    return self.currentContext().callWithContext(ctx, func, *args, **kw)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
       	    return func(*args,**kw)
       	--- <exception caught here> ---
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/twisted/internet/selectreactor.py", line 150, in _doReadOrWrite
       	    why = getattr(selectable, method)()
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/twisted/internet/tcp.py", line 203, in doRead
       	    return self._dataReceived(data)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/twisted/internet/tcp.py", line 209, in _dataReceived
       	    rval = self.protocol.dataReceived(data)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/twisted/internet/endpoints.py", line 85, in dataReceived
       	    return self._wrappedProtocol.dataReceived(data)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/ldaptor/protocols/ldap/ldapclient.py", line 54, in dataReceived
       	    o, bytes = pureber.berDecodeObject(self.berdecoder, self.buffer)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/ldaptor/protocols/pureber.py", line 373, in berDecodeObject
       	    berdecoder=inh)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/ldaptor/protocols/pureldap.py", line 57, in fromBER
       	    l = berDecodeMultiple(content, berdecoder)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/ldaptor/protocols/pureber.py", line 392, in berDecodeMultiple
       	    n, bytes = berDecodeObject(berdecoder, content)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/ldaptor/protocols/pureber.py", line 373, in berDecodeObject
       	    berdecoder=inh)
       	  File "/Users/i_m/projects/current/ldapcached/.venv/lib/python2.7/site-packages/ldaptor/protocols/pureldap.py", line 1293, in fromBER
       	    assert 3<=len(l)<=4
       	exceptions.AssertionError:

2016-12-02 05:09:10+0300 [LDAPClient,client] [ERROR] Could not connect to proxied server.  Error was:
       	[Failure instance: Traceback (failure with no frames): <type 'exceptions.AssertionError'>:
       	]
2016-12-02 05:09:10+0300 [LDAPClient,client] Stopping factory <ldaptor.protocols.ldap.ldapconnector.OneShotFactory instance at 0x109125bd8>
```
This happens because current code expects maximum of 4 encoded values in extended response, but up to 6 may actually exists:
```
        LDAPResult ::= SEQUENCE {
             resultCode         ENUMERATED {
                  ...  },
             matchedDN          LDAPDN,
             diagnosticMessage  LDAPString,
             referral           [3] Referral OPTIONAL }

        ExtendedResponse ::= [APPLICATION 24] SEQUENCE {
             COMPONENTS OF LDAPResult,
             responseName     [10] LDAPOID OPTIONAL,
             responseValue    [11] OCTET STRING OPTIONAL }
```
This PR fixes my problem and also adds additional check for STARTTLS response.

### DIFF
--- a/ldaptor/protocols/ldap/ldapclient.py
+++ b/ldaptor/protocols/ldap/ldapclient.py
@@ -33,6 +33,14 @@ class LDAPStartTLSBusyError(ldaperrors.LDAPOperationsError):
     def __str__(self):
         return 'Cannot STARTTLS while operations on wire: %r' % self.onwire
 
+class LDAPStartTLSInvalidResponseName(ldaperrors.LDAPException):
+    def __init__(self, responseName):
+        self.responseName = responseName
+        ldaperrors.LDAPException.__init__(self)
+
+    def __str__(self):
+        return 'Invalid responseName in STARTTLS response: %r' % (self.responseName, )
+
 class LDAPClient(protocol.Protocol):
     """An LDAP client"""
     debug = False
@@ -205,6 +213,10 @@ class LDAPClient(protocol.Protocol):
         assert msg.referral is None #TODO
         if msg.resultCode!=ldaperrors.Success.resultCode:
             raise ldaperrors.get(msg.resultCode, msg.errorMessage)
+
+        if (msg.responseName is not None) and \
+                (msg.responseName != pureldap.LDAPStartTLSResponse.oid):
+            raise LDAPStartTLSInvalidResponseName(msg.responseName)
 
         self.transport.startTLS(ctx)
         return self
--- a/ldaptor/protocols/pureldap.py
+++ b/ldaptor/protocols/pureldap.py
@@ -1338,17 +1338,29 @@ class LDAPExtendedResponse(LDAPResult):
         l = berDecodeMultiple(content, LDAPBERDecoderContext_LDAPExtendedResponse(
             fallback=berdecoder))
 
-        assert 3<=len(l)<=4
+        assert 3<=len(l)<=6
 
         referral = None
-        #if (l[3:] and isinstance(l[3], LDAPReferral)):
-            #TODO support referrals
-            #self.referral=self.data[0]
+        responseName = None
+        response = None
+        for obj in l[3:]:
+            if isinstance(obj, LDAPResponseName):
+                responseName = obj.value
+            elif isinstance(obj, LDAPResponse):
+                response = obj.value
+            elif isinstance(obj, LDAPReferral):
+                #TODO support referrals
+                #self.referral=self.data[0]
+                pass
+            else:
+                assert False
 
         r = klass(resultCode=l[0].value,
                   matchedDN=l[1].value,
                   errorMessage=l[2].value,
                   referral=referral,
+                  responseName=responseName,
+                  response=response,
                   tag=tag)
         return r
     fromBER = classmethod(fromBER)
--- a/ldaptor/test/test_pureldap.py
+++ b/ldaptor/test/test_pureldap.py
@@ -561,6 +561,38 @@ class KnownValues(unittest.TestCase):
             + l('42.42.42'))
          ),
 
+        (pureldap.LDAPExtendedResponse,
+         [],
+         {'resultCode': 49,
+          'matchedDN': 'foo',
+          'errorMessage': 'bar',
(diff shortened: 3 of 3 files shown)


psi29a:
I like this PR very much! 

However , it looks like this broke Python 2.6, any way to fix this?

cwaldbieser:
Looks like it is just complaining about the placeholder not having an explicit position number in a format specifier.  If you just throw in a '0' or two, it should pass on Python v2.6.

im-0:
It seems that problem with python 2.6 is caused by some external library, not by my code.

psi29a:
I problem exists here: 
https://github.com/hynek/attrs/blob/master/src/attr/_make.py

@hynek is the author of  attr, but according to its pypi page it only supports python 2.7

I might be able to send a PR that fixes these issues and see if he accepts it. That will have to be tomorrow-ish unless someone beats me to it. :)

hynek:
I’m sorry but the lack of attrs’ support of Python 2.6 is very much intentional and I may even add a way to break it on import to avoid subtle breakage.

Both Django and Twisted dropped 2.6 a while ago.  There’s no reason for little guys to support enterprise users for free forever.  2.6. is from 2008 and isn’t supported by core dev since 2013. http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html http://www.snarky.ca/stop-using-python-2-6

cwaldbieser:
@psi29a - What do you think?  Dropping support for Python 2.6 makes sense, since Twisted core only supports Python 2.7.  I'd be OK with dropping the v2.6 support.

psi29a:
Thanks for the heads up @hynek.  I'm fine with dropping Python 2.6, we're not using it at my work and no one else has complained. If they want to use the latest Twisted then they will have to be 2.7 anyway.

I'll merge this in then and go about purging 2.6 support. ;)

psi29a:
@im-0 thank you very much for the PR!

PR #68, opened 2016-12-08
Don't add callbacks for each child individually.

anlu:
This fixes an issue when there are a large number of children and
the original callback has been called, quickly causing the maximum
recursive limit to be exceeded since each additional _processOneChild callback gets evaluated immediately by Twisted.

### DIFF
--- a/ldaptor/entryhelpers.py
+++ b/ldaptor/entryhelpers.py
@@ -129,15 +129,10 @@ class SubtreeFromChildrenMixin(object):
         else:
             callback(self)
             d = self.children()
-            def _processOneChild(_, children, callback):
-                if not children:
-                    return None
-
-                c = children.pop()
-                d = c.subtree(callback)
-                d.addCallback(_processOneChild, children, callback)
             def _gotChildren(children, callback):
-                _processOneChild(None, children, callback)
+                if children:
+                    while len(children) > 0:
+                        children.pop().subtree(callback)
             d.addCallback(_gotChildren, callback)
             return d
 


cwaldbieser:
Is there some reason to use `while` ... `.pop()` instead of the more typical `for c in children:`?

anlu:
Yes, the old version also popped from the end of the list, and there were a few tests that explicitly specify the order to pass. If someone can confirm that there isn't a practical reason why the order should matter (I am very new to LDAP), I'm happy to change it to the more idiomatic `for c in children:`.

cwaldbieser:
`for c in children` will process the list items in the same order.  From the Python docs[1]:

> Python’s for statement iterates over the items of any sequence (a list or a string), **in the order** that they appear in the sequence.

(emphasis is mine).

Under normal conditions, the order that results are returned from an LDAP search are undefined.  I don't have the RFC reference for this, though.  It is just an observation I have made.

[1] https://docs.python.org/2/tutorial/controlflow.html#for-statements

cwaldbieser:
Found it in RFC4511 [1]:

> The SearchResultEntry and SearchResultReference messages may come in any order.

[1] https://tools.ietf.org/html/rfc4511#section-4.5.2

anlu:
I think we're talking past each other here...for c in children will go in order but pop will go in reverse order:
```
>>> children = [1, 2, 3]
>>> for c in children:
...     print c
...
1
2
3
>>> while len(children) > 0:
...     print children.pop()
...
3
2
1
```

I think you're thinking about pop(0):
```
>>> children = [1, 2, 3]
>>> while len(children) > 0:
...     print children.pop(0)
...
1
2
3
```

Regardless, I will update to `for c in children:` and update the tests.

cwaldbieser:
I ran some functional tests and everything looks good.  Thanks for contributing!

psi29a:
Thank you @anlu and @cwaldbieser

PR #69, opened 2016-12-20
Use a dictionary for children of ReadOnlyInMemoryLDAPEntry

anlu:
Using a dictionary of RDNs to children Entries dramatically speeds up adding and deleting entries since we don't have to iterate over an entire list of children per entry.


psi29a:
Thank you for this and for the test updates. It looks OK to me.  

@cwaldbieser What do you think? :)

cwaldbieser:
Looks good to me.  I ran it against the quick-start server and it seems to behave normally.  Code changes look good. I've merged the PR.

Thanks for contributing!

PR #72, opened 2016-12-28
test_server: Use list and assertItemsEqual to compare ldap results

tonich-sh:


cwaldbieser:
On your line 673, I see `self.assertItemEqual`, but I couldn't find a reference to this method in the `unittest` module.  Is it supposed to be another case of the plural `assertItemsEqual()` method?  And the next argument was changed from a list to a dictionary-- I'm not sure I'm following what you want to change in that test.  The fact that this test is a TODO makes the error fall under the radar.

tonich-sh:
The line 673 is not my change. It's just a typo IMO. This test does'n reach this code now. And the next argument is not a dictionary - this is an another form of creating a set instance. I just removed usage of deprecated "sets" module.

cwaldbieser:
Thanks for correcting this.  I merged it.  If you can update your other PR with this change, I am hoping that will pass, too.

tonich-sh:
I don't see a merging commit. Maybe you forgot to push?

cwaldbieser:
Sorry I missed that!  That's what I get for trying to merge over the holidays!

psi29a:
Thanks for following this up, both of you. :)

PR #74, opened 2017-07-14
Added custom escaping functionality when generating LDAP filter strings

### LINKED ISSUE #73: Allow custom escaping functions when generating LDAP filters
Certain LDAP implementations have different escaping requirements for various attributes. For example, in Active Directory the `objectGUID` attribute must have its value in hexadecimal representation. Here is a concise description of the issue: https://support.pingidentity.com/LDAP-Filter-on-ObjectGUID

In `ldaptor` it currently looks something like this:

```python
#!/usr/bin/env python

from ldaptor.protocols import pureldap


filt = pureldap.LDAPFilter_equalityMatch(
    attributeDesc=pureldap.LDAPAttributeDescription('objectGUID'),
    assertionValue=pureldap.LDAPAttributeValue('90395F191AB51B4A9E9686C66CB18D11')
)

print filt.asText()
```

And we get the following result:

```
$ python escape.py 
(objectGUID=90395F191AB51B4A9E9686C66CB18D11)
```

Since this value is automatically escaped there's no opportunity to perform the hexadecimal escaping. If we try that beforehand the value will be double escaped. My question is, do you know of any good way to perform some kind of custom escaping of an attribute value? If not, I'd like to purpose a change that would allow for this.

I purpose all `LDAPFilter_*` classes take an optional argument when initialized that allows for custom escaping. It would look something like this:

```python
#!/usr/bin/env python

from ldaptor.protocols import pureldap


filt = pureldap.LDAPFilter_equalityMatch(
    attributeDesc=pureldap.LDAPAttributeDescription('objectGUID'),
    assertionValue=pureldap.LDAPAttributeValue('90395F191AB51B4A9E9686C66CB18D11'),
    escaper=OBJECTGUID_ESCAPE_FUNCTION
)

print filt.asText()
```

Which would produce:

```
$ python escape.py 
(objectGUID=\90\39\5F\19\1A\B5\1B\4A\9E\96\86\C6\6C\B1\8D\11)
```

But, `escaper` could be any escaping function. In this scenario we've just defined some `OBJECTGUID_ESCAPE_FUNCTION` function that does what we need for `objectGUID`. Finally, if no `escaper` function is specified we simply fallback on the original behavior.

I was originally planning on simply submitting a PR with the intended changes, but I made the issue first in case you were aware of some way to do this without any changes to the code base. If that's not the case I'll get to work on submitting a PR :)

Let me know if you have any concerns!

mschwager:
Fixes #73, see issue for further details.

### DIFF
--- a/ldaptor/protocols/pureber.py
+++ b/ldaptor/protocols/pureber.py
@@ -113,7 +113,7 @@ def ber2int(e, signed=True):
         v=(v<<8) | ord(e[i])
     return v
 
-class BERBase:
+class BERBase(object):
     tag = None
 
     def identification(self):
--- a/ldaptor/protocols/pureldap.py
+++ b/ldaptor/protocols/pureldap.py
@@ -15,6 +15,8 @@
 
 """LDAP protocol message conversion; no application logic here."""
 
+import string
+
 from pureber import (
 
     BERBoolean, BERDecoderContext, BEREnumerated, BERInteger, BERNull,
@@ -40,11 +42,23 @@ def escape(s):
     s = s.replace('\0', r'\00')
     return s
 
+def binary_escape(s):
+    return ''.join('\\{0:02x}'.format(ord(c)) for c in s)
+
+def smart_escape(s, threshold=0.30):
+    binary_count = sum(c not in string.printable for c in s)
+    if float(binary_count) / float(len(s)) > threshold:
+        return binary_escape(s)
+
+    return escape(s)
+
 class LDAPInteger(BERInteger):
     pass
 
 class LDAPString(BEROctetString):
-    pass
+    def __init__(self, *args, **kwargs):
+        self.escaper = kwargs.pop('escaper', escape)
+        super(LDAPString, self).__init__(*args, **kwargs)
 
 class LDAPAttributeValue(BEROctetString):
     pass
@@ -346,11 +360,12 @@ class LDAPAttributeValueAssertion(BERSequence):
         return r
     fromBER = classmethod(fromBER)
 
-    def __init__(self, attributeDesc=None, assertionValue=None, tag=None):
+    def __init__(self, attributeDesc=None, assertionValue=None, tag=None, escaper=escape):
         BERSequence.__init__(self, value=[], tag=tag)
         assert attributeDesc is not None
         self.attributeDesc=attributeDesc
         self.assertionValue=assertionValue
+        self.escaper=escaper
 
     def __str__(self):
         return str(BERSequence([self.attributeDesc,
@@ -428,26 +443,26 @@ class LDAPFilter_equalityMatch(LDAPAttributeValueAssertion):
 
     def asText(self):
         return '('+self.attributeDesc.value+'=' \
-               +escape(self.assertionValue.value)+')'
+               +self.escaper(self.assertionValue.value)+')'
 
 class LDAPFilter_substrings_initial(LDAPString):
     tag = CLASS_CONTEXT|0x00
 
     def asText(self):
-        return escape(self.value)
+        return self.escaper(self.value)
 
 
 class LDAPFilter_substrings_any(LDAPString):
     tag = CLASS_CONTEXT|0x01
 
     def asText(self):
-        return escape(self.value)
+        return self.escaper(self.value)
 
 class LDAPFilter_substrings_final(LDAPString):
     tag = CLASS_CONTEXT|0x02
 
     def asText(self):
-        return escape(self.value)
+        return self.escaper(self.value)
 
 class LDAPBERDecoderContext_Filter_substrings(BERDecoderContext):
     Identities = {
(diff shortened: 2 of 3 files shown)


ldaptor/protocols/pureber.py
```diff
@@ -113,7 +113,7 @@ def ber2int(e, signed=True):
         v=(v<<8) | ord(e[i])
     return v
 
-class BERBase:
+class BERBase(object):
```
mschwager: This change is required or else the `super(LDAPString, self)` call will fail.

psi29a:
Wow, thank you for not  only improving Ldaptor but also adding test-coverage for the feature as well! 

Could you also add an entry in the the docs/source/NEWS.rst file?  A 17.1 (UNRELEASED) entry would be fine. :)

As for the rest, it looks good. @cwaldbieser what do you think?

cwaldbieser:
The changes make sense to me at a programmatic level-- I understand how adding the option to provide your own escape function allows you to customize the various `pureldap.LDAPFilter_*` classes.  And the tests are a welcome addition.  I guess I am a bit confused as to *how* the changes would be used.  I mostly think of escaping filters as something that needs to be done from the client side, and I don't see the classes used in `ldapclient.LDAPClient`, for example.

Is the idea that you want to make an LDAP server that expects the filter it receives to be escaped that way?  Or am I just overthinking things?

cwaldbieser:
I reviewed the actual flow a little more in depth, and I can see what I was not understanding.  If we use the quickstart LDAP client[1] as an example, we see the filter, as a python string, is passed to `ldapsyntax.LDAPEntry.search()`.  This is internally converted by `ldapfilter.parseFilter()` into a `pureldap.LDAPFilter_equalityMatch` or some other match class(es) depending on how the string is parsed.

What strikes me as odd is that the various `pureldap.LDAPFilter_*` classes are escaping anything at all.  I had assumed up until this point that it was totally up to the client to perform the escaping.  Now that I understand what's going on, I should be able to tell if the code changes make sense at a semantic level.

[1] http://ldaptor.readthedocs.io/en/latest/quickstart.html

cwaldbieser:
One thing I am not seeing is how the `asText()` methods on the various filter objects translate into on-the-wire representations.  I'll have another look in the morning.

cwaldbieser:
I think I understand what is really going on now, and I think it bears a bit more discussion in the [issue](https://github.com/twisted/ldaptor/issues/73).

The realization here is that the changes allow for a custom escape function for displaying the textual representation of a filter, but the textual representation is really only used in debugging-- not for actually passing the filter to an LDAP server.  I'll explain more in the issue.

mschwager:
Hey all, thanks for responding!

@psi29a if we deem this change worthwhile I'll definitely update the `NEWS` file.

@cwaldbieser Hmm, interesting notes. The way we're currently using the `ldaptor` library we pass filter strings between a client and server in order to perform the LDAP search. In other words, we have a server that passes LDAP filter queries to a client installed on a machine which speaks to the LDAP store. This may not be the most optimal way to perform the search, but we either end up using the `asText` function, or string manipulation to get the filter we want.

An alternative solution here is to remove the automatic `escape` call from the `asText` functions. However, this could have larger downstream implications if people are depending on this behavior.

Either way, let me know what you think! :)

cwaldbieser:
@mschwager - I think your idea makes sense.  There are multiple textual representations which are equivalent, but circumstances may dictate which one is more appropriate for your particular needs.  That said, dealing with GUIDs in AD is likely to be fairly common, and I would not be against making the default representation more in line with what one might reasonably expected.

I created a (gist)[https://gist.github.com/cwaldbieser/fe356d4eacbea717c4052ad2664d94b6] with some code from (stack overflow)[https://stackoverflow.com/questions/1446549/how-to-identify-binary-and-text-files-using-python] that I think might fit nicely into your PR.  I still think it should still be possible to customize the textual representation however one might want.  But I think the default behavior should maybe be less surprising.

Does that make sense?

mschwager:
That makes sense to me. However, I still have concerns for downstream code that may be relying on the current `escape` implementation. Since we're including the possibility for arbitrary escaping it makes sense to me to just add an additional function that can be used in the `escaper` argument. Something like:

```python
import string

def escape(s):
    s = s.replace('\\', r'\5c')
    s = s.replace('*', r'\2a')
    s = s.replace('(', r'\28')
    s = s.replace(')', r'\29')
    s = s.replace('\0', r'\00')
    return s

def binary_escape(s):
    return ''.join('\\{0:02x}'.format(ord(c)) for c in s)

def smart_escape(s, threshold=0.30):
    binary_count = sum(c not in string.printable for c in s)
    if float(binary_count) / float(len(s)) > threshold:
        return binary_escape(s)

    return escape(s)
```

Then people can use the `smart_escape` function as they see fit, while we still maintain backwards compatibility. Although, maybe I'm over-estimating the effects of changing the `escape` function behavior. Thoughts?

cwaldbieser:
That seems reasonable to me.  Since the `escape` function is more or less public, some users may simply be using it for client-side escaping.  `smart_escape` would be valid for that case, but `escape` may be preferable if you are never going to actually view the escaped assertion.

I don't want to hold up your PR, but there is one more related issue I though of-- it seems like even with the customizable `escape` option on the various `LDAPAttributeValueAssertion` subclasses, you basically need to construct the filter from objects if you want to make use of custom escaping.  Using `ldapfilter.parseFilter()` either directly or indirectly uses the default escaping.  I wonder if there might be some way to register a default escaping function.  I haven't given too much thought to how it might be implemented, but that might be a nice touch.

If you're not interested, I'll accept your PR with just the option to include custom escape functions.

cwaldbieser:
FYI - I did verify that I could query an account by objectGUID using the current version of Ldaptor by including escaped binary data in the filter string.  That's not how you are using the library, but it will be a good reminder to me at some point that this change is about human-readable representation and not about escaping data sent to an LDAP service.  That still needs to happen at the client.

mschwager:
@cwaldbieser Thanks for all the feedback! I've added the `smart_escape` functionality we've been discussing. As far as `ldapfilter.parseFilter` goes, I'd be okay with adding escaping functionality to that portion of the codebase as well, but I'd like to do it in a separate PR. I think this one is good to go :+1:

cwaldbieser:
Thanks for contributing!

psi29a:
Thank you for this! :)

PR #77, opened 2017-09-05
Added ability to logically compare LDAPFilterSet objects

### LINKED ISSUE #76: Allow for logical comparisons in LDAPFilterSet
Hey again!

I'm back with some more LDAP filter requests. The basic jist of this request is, I'd like to be able to perform logical comparisons on LDAP filter AND and OR statements. For example, consider the following:

```
(&(foo=1)(bar=2)) == (&(bar=2)(foo=1))
```

I would expect to be true. Note that this may not be an identical comparison in terms of performance. E.g.

```
if long_comparison_operation() and short_comparison_operation():
    ...
```

This would be logically equivalent to the following statement, but not necessarily the same in terms of performance:

```
if short_comparison_operation() and long_comparison_operation():
    ...
```

We can avoid making the second comparison operation if the first one is false, and thus avoid the long comparison operation.

In terms of actual behavior within this codebase, I'd like to see something like the following work:

```python
from ldaptor.protocols import pureldap

filter1 = pureldap.LDAPFilter_and([
    pureldap.LDAPFilter_equalityMatch(
        attributeDesc=pureldap.LDAPAttributeDescription('foo'),
        assertionValue=pureldap.LDAPAttributeValue('1')
    ),
    pureldap.LDAPFilter_equalityMatch(
        attributeDesc=pureldap.LDAPAttributeDescription('bar'),
        assertionValue=pureldap.LDAPAttributeValue('2')
    ),
])
filter2 = pureldap.LDAPFilter_and([
    pureldap.LDAPFilter_equalityMatch(
        attributeDesc=pureldap.LDAPAttributeDescription('bar'),
        assertionValue=pureldap.LDAPAttributeValue('2')
    ),
    pureldap.LDAPFilter_equalityMatch(
        attributeDesc=pureldap.LDAPAttributeDescription('foo'),
        assertionValue=pureldap.LDAPAttributeValue('1')
    ),
])
                                                                
assert filter1 == filter2
```

Why would we want to perform comparisons like this? We could imagine some unordered LDAP filter generation strategy where we'd like to confirm two filters are equivalent. For example, consider a Python `dict` (unordered `key: value` pairs) that is eventually converted to an LDAP filter. We may want to confirm that LDAP filters are logically equivalent, but not necessarily the exact same string or `LDAPFilter_*` object. E.g.

```python
filter1 = {
    'attribute1': 'value1',
    'attribute2': 'value2',
}

filter2 = {
    'attribute1': 'value1',
    'attribute2': 'value2',
}
```

We're not guaranteed an ordering if we do something like `filter1.items()`, but it's conceivable an LDAP filter could be generated from this `dict`. Further, it's conceivable we'd like to confirm two of these LDAP filters generated from unordered `dict`s are equivalent.

I have a PR in mind for implementing this behavior, but I opened this issue to A.) ensure this is something we should be considering and B.) open the door for other options than my PR implementation.

Let me know what you think!

mschwager:
Fixes #76 

Here's my purposed fix for the issue I was having in #76. I like it because it's a fairly simple solution to the problem at hand.

It's worth noting, I did consider an alternate solution. Namely, using the `collections.Counter` object. I would've actually preferred this solution because it's faster, but, unfortunately it doesn't work for nested OR and AND filters. E.g. `test_nested_equal`. Something like:

```diff
diff --git a/ldaptor/protocols/pureber.py b/ldaptor/protocols/pureber.py
index 7b12678..9052e13 100644
--- a/ldaptor/protocols/pureber.py
+++ b/ldaptor/protocols/pureber.py
@@ -305,6 +305,9 @@ class BERSequence(BERStructured, UserList.UserList):
             return self.__class__.__name__+"(value=%s, tag=%d)" \
                    %(repr(self.data), self.tag)
 
+    def __hash__(self):
+        return hash(self.asText())
+
 
 class BERSequenceOf(BERSequence):
     pass
diff --git a/ldaptor/protocols/pureldap.py b/ldaptor/protocols/pureldap.py
index f97973c..ef90fcf 100644
--- a/ldaptor/protocols/pureldap.py
+++ b/ldaptor/protocols/pureldap.py
@@ -15,7 +15,7 @@
 
 """LDAP protocol message conversion; no application logic here."""
 
-import operator
+import collections
 import string
 
 from pureber import (
@@ -394,8 +394,10 @@ class LDAPFilterSet(BERSet):
     fromBER = classmethod(fromBER)
 
     def __eq__(self, rhs):
-        return (sorted(self, key=operator.methodcaller('asText')) ==
-            sorted(rhs, key=operator.methodcaller('asText')))
+        return (
+            collections.Counter(self) ==
+            collections.Counter(rhs)
+        )
 
 class LDAPFilter_and(LDAPFilterSet):
     tag = CLASS_CONTEXT|0x00
```

`Counter` uses hashing on the backend, so we'd have to implement `__hash__`, and is preferable because it runs in `O(n)`, but fails at recursive comparisons. `sorted` runs in `O(n log n)`, and works for recursive comparisons because to sort we will recursively be calling `__eq__`.

Note that `Counter` is essentially what [assertItemsEqual](https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.assertItemsEqual) does, but alas won't work recursively :(

Let me know what you think! Especially if you can see some way to make `Counter` work!

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+Release 17.1 (UNRELEASED)
+-------------------------
+
+Features
+^^^^^^^^
+
+- Ability to logically compare ldaptor.protocols.pureldap.LDAPFilter_and and ldaptor.protocols.pureldap.LDAPFilter_or objects with ==.
+- Ability to customize ldaptor.protocols.pureldap.LDAPFilter_* object's encoding of values when using asText.
+
 Release 16.0 (2016-06-07)
 -------------------------
 
--- a/ldaptor/protocols/pureldap.py
+++ b/ldaptor/protocols/pureldap.py
@@ -392,6 +392,15 @@ class LDAPFilterSet(BERSet):
         return r
     fromBER = classmethod(fromBER)
 
+    def __eq__(self, rhs):
+        # Fast paths
+        if self is rhs:
+            return True
+        elif len(self) != len(rhs):
+            return False
+
+        return sorted(self, key=str) == sorted(rhs, key=str)
+
 class LDAPFilter_and(LDAPFilterSet):
     tag = CLASS_CONTEXT|0x00
 
--- a/ldaptor/test/test_pureldap.py
+++ b/ldaptor/test/test_pureldap.py
@@ -868,3 +868,173 @@ class TestEscaping(unittest.TestCase):
         for filt, expected in filters:
             result = filt.asText()
             self.assertEqual(expected, result)
+
+class TestFilterSetEquality(unittest.TestCase):
+    def test_basic_and_equal(self):
+        filter1 = pureldap.LDAPFilter_and([
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('foo'),
+                assertionValue=pureldap.LDAPAttributeValue('1')
+            ),
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('bar'),
+                assertionValue=pureldap.LDAPAttributeValue('2')
+            ),
+        ])
+        filter2 = pureldap.LDAPFilter_and([
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('bar'),
+                assertionValue=pureldap.LDAPAttributeValue('2')
+            ),
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('foo'),
+                assertionValue=pureldap.LDAPAttributeValue('1')
+            ),
+        ])
+
+        self.assertEqual(filter1, filter2)
+
+    def test_basic_and_not_equal(self):
+        filter1 = pureldap.LDAPFilter_and([
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('foo'),
+                assertionValue=pureldap.LDAPAttributeValue('1')
+            ),
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('bar'),
+                assertionValue=pureldap.LDAPAttributeValue('2')
+            ),
+        ])
+        filter2 = pureldap.LDAPFilter_and([
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('bar'),
+                assertionValue=pureldap.LDAPAttributeValue('1')
+            ),
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('foo'),
+                assertionValue=pureldap.LDAPAttributeValue('1')
+            ),
+        ])
+
+        self.assertNotEqual(filter1, filter2)
+
+    def test_basic_or_equal(self):
+        filter1 = pureldap.LDAPFilter_or([
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('foo'),
+                assertionValue=pureldap.LDAPAttributeValue('1')
+            ),
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('bar'),
+                assertionValue=pureldap.LDAPAttributeValue('2')
+            ),
+        ])
+        filter2 = pureldap.LDAPFilter_or([
+            pureldap.LDAPFilter_equalityMatch(
+                attributeDesc=pureldap.LDAPAttributeDescription('bar'),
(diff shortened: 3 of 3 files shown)


psi29a:
Thanks for creating this MR, disregard the pypy errors, seems we need to update this.

:1st_place_medal: for the tests :)

mschwager:
Ahh, I see. Is there anything I can do to help move forward with this?

cwaldbieser:
One simple optimization might be to check if the identity if the left and right hand sides are equal.  E.g. if `id(self) == id(rhs)`, then the 2 items must be equal.  Not sure how much this will help in practice.

cwaldbieser:
This is a bit pathological, but by using `asText()` and implicitly using the `self.escaper()`, couldn't 2 filters that are equal end up not comparing the same?  E.g. if I have minimal escaping on one filter but complete filtering on another, their textual representations would be different.

mschwager:
Thanks for the feedback.

Per the optimizations, typically I like to keep the code as simple as possible, then tackle optimizations when needed. With that being said, though, doing an identity check first doesn't complicate things too much. Further we could check that their `len` is the same before sorting as well :+1:

As for using `asText`, this point is particularly interesting. At first I was inclined to agree with you that two filters that are otherwise the same, but have differing escape functions, should be considered equal, and the code may not currently represent that. But now I'm leaning the other way. For example, if we have different escape functions we could get the following filter strings from the same object: `(attribute=A)` and `(attribute=01000001)`. Now I believe that these should be considered different. The LDAP directory *should* return the same result if it supports different encodings for the same query, however, say the directory requires the attribute value to be binary encoded. Then the first query would fail and the second would succeed, which fundamentally means the filter objects are different.

Thoughts?

cwaldbieser:
Q1: Will the directory service return identical results for 2 filters.
Q2: Are 2 filters semantically identical.

Questions 1 and 2 are different questions.  Sometimes this is subtle, sometimes not.  For example, take  `(name=a*)` and `(name=A*)`.  These filters have a difference in casing, but depending on the schema rules for "name" at the directory, the filters could be equivalent.

I think the best we can hope to do is to answer Q2.  Further, I think we need to assume that [logical equivalences](https://en.wikipedia.org/wiki/Logical_equivalence) are the only differences we could really look at.

Your use case may be even simpler-- I think you just want to see if a filter is equivalent in the case that the terms of a binary AND or OR operation are switched.

I'm not sure if you are interested in something like De Morgan's laws.  E.g. `(!(|(givenName=Carl)(sn=Waldbieser)))` == `(&(!(givenName=Carl))(!(sn=Waldbieser)))` .  Both filters eliminate entries with givenName "Carl" AND entries with sn "Waldbieser".  They are logically equivalent filters, but the forms are rather distinct.

It makes me wonder if overloading the __eq__ operator is the correct kind of comparison.  Maybe a function/method like "test_logical_equivalence()" or a name more appropriate to whatever kind of comparison we are making might be more appropriate.

mschwager:
I like your idea of not concerning ourselves with what a directory will consider an equivalent filter. It simplifies much of this, and makes sense since this library shouldn't be trying to anticipate the needs of many different directories.

With that being said, I agree that it would then make sense to avoid the escaping function when considering equivalence. We'll only consider the attribute and value when checking for equivalence and not the escaped/encoded value. I'll make that change.

Further, you are also correct that I'm only trying to find equivalences between simple AND and ORs, not complex logical equivalents, e.g. DeMorgan's. I'll save solving [boolean satisfiability](https://en.wikipedia.org/wiki/Boolean_satisfiability_problem) for another day :)

To your last point, I do particularly like using `__eq__`. It makes comparison Just Work™ with different libraries and operations. For example, [assertEqual](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertEqual) and other functions that use `==` will just work without having to add special comparators.

I did briefly look into why different escapers were resulting in the same filter. I would've expected otherwise given the predicate `asText`. I did come up with a test to prove your point that there could be incorrect results, I'll add it to the PR. To remedy the issue I've changed to using `str` instead of `asText`. Good catch!

So I've:

* Added some comparison optimizations
* Changed the semantics to not rely on `asText` and thus not rely on the escaper
* Kept `__eq__`, I fairly partial to using it, let me know if this is a dealbreak

Let me know what you think!

cwaldbieser:
@psi29a - What do you think?  Operator overloading kind of rubs me the wrong way in this case, but maybe I'm just being too picky?  Otherwise, the PR looks OK to me.

psi29a:
Honestly, I think it is a elegant way of handling things, it is self-contained in the overload and there are test-coverage enough that we should be able to track it going forward for regressions.

psi29a:
Please add an:
Release 17.1 (UNRELEASED)
section in our NEWS.rst and add your bug-fix.
https://github.com/twisted/ldaptor/blob/master/docs/source/NEWS.rst

After that, we're ready to merge. :)

(Not dumping this on you, I'm asking that any PRs that come in, to add an entry when fixing bugs, adding features or changing the API. That way we don't have to go hunting when making a release.)

mschwager:
No sweat, I'm a big fan of maintaining a CHANGELOG. I added this feature to NEWS and also added the functionality I missed from my last PR :)

psi29a:
Merging! Thanks again.

Issue #78, opened 2017-09-22
Allow BIND using UPN to fake a AD LDAP server.


adiroiban:
Not sure if someone can find it useful.

Active Directory LDAP implementation allows to bind using UPN (user@domain.com) or down-level logon name (user\domain)

I am using ldaptor to support the testing for an ldap client code.

For some unit / integration tests for me is easier to use ldaptor, so I have created this modified server which allows to bind using something similar to an UPN 

Code is here

https://gist.github.com/adiroiban/cc60a78e0275e07869e3ed1fb0114621

If there is interest, I can look at pushing a PR... but the code is not designed for production :)

cwaldbieser:
I'm not sure if it belongs in the code base-- but it might be interesting to put it in the [cookbook under servers.](http://ldaptor.readthedocs.io/en/latest/cookbook/servers.html) .  Could you write it up as a recipe?

adiroiban:
Fixed in #79

PR #79, opened 2017-09-26
Add example for AD like BIND.

adiroiban:
Scope
=====

See the discussion from #78 

Changes
=======

I have moved all examples into a single place.

I chose that single place to be the Sphinx source so that it is easy to reference them from the source documentation.

I prefer `literalinclude` as I find it easier to manage and in this way is easier to test the examples.

I have converted the examples into a module to make it easier to import it during tests.

For the tests, I am messing with the sys.path.

I don't know if trial has something like tearDownModule ... so I don't know how to do the cleanup.
I hope that there is no harm in this.

As a drive-by, I have creates a tox env for building the documentation so that we can have a canonical name to create the docs.

As a drive-by I added the travis-ci badge to README as it makes it easier to check if master is green.

How to check
==========

Let me know if you are ok with the drive-by changes. Not sure why the examples were outside of docs/source in the first place :)

I wrote the test, so I hope that in this way the example will stay up to date.


adiroiban:
should be ready for review. I see that master is failing on pypi

needs-review

psi29a:
Fixed pypy (by removing it for now)

cwaldbieser:
Made a couple small edits-- typo in "bobby", altered phrasing in a few comments.  It was a good example of how you can intercept an LDAP request and adapt it to features of the back end.  

Thanks, for contributing!

adiroiban:
Thanks for the review and thanks for keeping ldaptor alive :)

PR #81, opened 2017-10-04
Use classmethod decorator instead of function

mschwager:
Hey all,

I noticed some files in this project are still using the old `classmethod` function instead of the new-style decorator. This project must go back pretty far to be using the original function! Sure enough, I checked the log and the first commit was July 2002! Looks like the decorators went in with [PEP 318](https://www.python.org/dev/peps/pep-0318/) in Python 2.4 in June 2003!

Based on your `.travis.yml` it seems you're only supporting Python 2.7, or at least it's safe to say you're supporting 2.4 and higher ;)

In this PR I've updated the code to use the `@classmethod` decorator.


psi29a:
Looks good to me, update the changelog and we can get this baby merged in. ;)

mschwager:
Great, done!

PR #82, opened 2017-10-18
made string comparision case insensitive by using lowercase function

crucidal:
Generally the userDN is case insensitive in the ldap servers I have seen so far.
For ldaptor this was not the case, so here's a pr that resolves that.

Let me know what you think!

Kind Regards,
Chris

### DIFF
--- a/ldaptor/__init__.py
+++ b/ldaptor/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "16.0.0"
+__version__ = "16.0.1"
 
 __title__ = "ldaptor"
 __description__ = "A Pure-Python Twisted library for LDAP"
--- a/ldaptor/__init__.py
+++ b/ldaptor/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "16.0.1"
+__version__ = "16.0.0"
 
 __title__ = "ldaptor"
 __description__ = "A Pure-Python Twisted library for LDAP"
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -15,6 +15,11 @@ Changes
 
 - Using modern classmethod decorator instead of old-style method call.
 
+Bugfixes
+^^^^^^^^
+
+- DN matching now case insensitive.
+
 Release 16.0 (2016-06-07)
 -------------------------
 
--- a/ldaptor/protocols/ldap/distinguishedname.py
+++ b/ldaptor/protocols/ldap/distinguishedname.py
@@ -110,8 +110,8 @@ class LDAPAttributeTypeAndValue:
     def __eq__(self, other):
         if not isinstance(other, LDAPAttributeTypeAndValue):
             return NotImplemented
-        return (self.attributeType == other.attributeType
-                and self.value == other.value)
+        return (self.attributeType.lower() == other.attributeType.lower()
+                and self.value.lower() == other.value.lower())
 
     def __ne__(self, other):
         return not (self == other)


ldaptor/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "16.0.0"
+__version__ = "16.0.1"
```
psi29a: please don't bump the version number, that's out job. ;)

if we need to backport to a 16.0, then we'll handle that too. :)

crucidal: thank you for your feedback!

psi29a:
Looks good to me, you ok with this change @cwaldbieser ?
Can you also add an entry to the changelog please? :)

cwaldbieser:
I'm at a conference this week-- sorry if I'm slow to respond.

I believe that case sensitivity is supposed to follow the rules for the
individual components of a DN from their schema.  E.g. if the DN is made
from uid, ou, and o, and each part is case insensitive, (which I think they
are), the whole thing would be case sensitive.

I haven't had a good look at the commits-- if they are just for the built
in ldap servers, they don't really honor schemas anyway, so I'd probably be
ok with the changes.  If we are talking about changing the
DistinguishedName objects, I'd want to look more closely to make sure we're
not going to break anybody's code by matching when we shouldn't.

Thanks,
Carl


On Oct 18, 2017 1:30 PM, "Bret Curtis" <notifications@github.com> wrote:

Looks good to me, you ok with this change @cwaldbieser
<https://github.com/cwaldbieser> ?
Can you also add an entry to the changelog please? :)

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/twisted/ldaptor/pull/82#issuecomment-337718500>, or mute
the thread

psi29a:
All the documentation I've seen has referred to DNs as being case-insensative, but to be careful that some implementations don't honour this.

https://www.ietf.org/rfc/rfc2247.txt

>    The value of this attribute is a string holding one component of a
>    domain name.  The encoding of IA5String for use in LDAP is simply the
>    characters of the string itself.  The equality matching rule is case
>    insensitive, as is today's DNS.

http://ldapwiki.com/wiki/Distinguished%20Name%20Case%20Sensitivity

> During a bind Request, the Distinguished Name SHOULD NOT be case sensitive regardless of the makeup of the attributes (RDNs) within the Distinguished Name.

https://www.novell.com/documentation/developer/ndslib/schm_enu/data/sdk5582.html

> In NDAP, a Distinguished Name (DN) is transmitted on the wire as a Unicode string. The largest supported DN is 256 Unicode characters, not including the terminating NULL character. Distinguished Names are not case sensitive, even if one of the naming attributes is case sensitive.

https://www.ldap.com/ldap-dns-and-rdns

> Most LDAP client APIs offer some mechanism for comparing DNs (e.g., to determine whether they are equal, or to determine whether they are hierarchically related) or to generate some kind of "normalized" representation of a DN (e.g., with all extraneous spaces removed, with all case-insensitive characters converted to either all lowercase or all uppercase, etc.) that can be more safely compared using simple string processing.

https://www.drupal.org/node/1412108

> For most LDAP implementations this means, simply, distininguished names are case-insensitive.

crucidal:
@psi29a I updated the changelog and assumed that the changes have to be reflected in the latest unreleased version. (17.1 in this case) is that correct? :)

cwaldbieser:
OK, in light of Brett's fact-finding, I think this PR can be merged.  DNs should not consider case when determining if they match.

cwaldbieser:
Thanks for the PR!

Issue #55, opened 2016-06-20
Python3 support


Krejdom:
Hi, could you please support Python3? Many distributions are shifting to Python3 very soon and Python 2.7 will be supported only until 2020.

psi29a:
I have a python3 branch for about a year. Every month or so I check to see if the required twisted component has been ported yet or not. 

It is on the list. :)

s0undt3ch:
@psi29a: I noticed your py3 branch hasn't had an update for about a year now. With twsited 16.4.0 being released stating 35+ modules ported to py3, by any chance, the one you've been looking for is in that list?

psi29a:
I've been looking forward to that release a lot. When I have time, I'll give it a glance. I have a lot on my plate at the moment.  If you would like to take my branch for a spin and test with 16.4, let us know! :)

s0undt3ch:
I'll try it

adiroiban:
Is there a list of Twisted code which needs to be ported to py3?

We can use that to prioritize the Twisted port... and if there is a list in this ticket maybe people wanting to see ldaptor on py3 can help first with those ports.

Thanks :)

psi29a:
It's been awhile, I should give this a go again to see where we stand. I had a branch I was working on, but was blocked by 'srvconnect'.

My branch is here, 3 commits ahead of master.
https://github.com/psi29a/ldaptor/tree/python3

adiroiban:
The Twisted modules **not** ported to py3 are listed here https://github.com/twisted/twisted/blob/twisted-17.9.0/src/twisted/python/_setup.py#L366

I don't see twisted.names there so I guess that the whole DNS part was ported to py3

psi29a:
super...you can take my branch and move forward. I'll give it an eyeball over the weekend.

tardyp:
FWIW, we are willing to switch buildbot to ldaptor from ldap3 if this lib is compatible with py3.

adiroiban:
@psi29a thanks. Let me see what I can do... is only 3 comments but py3 porting stuff are mixed with style changes so it hard to tell which are the critical parts which needs extra care during review.

I will try to break it into 2 PRs. One for the cleanup and then another one for the actual py3 stuff.

I expect that the cleanup PR will be accepted without any controversy :)

s0undt3ch:
If this gets py3 support I'll also move an internal project to it(been using a java test LDAP server for integration tests, ugh)...

adiroiban:
[...]

PR #83, opened 2018-01-06
Style cleanup in preparation for python3 port.

adiroiban:
Scope
========

As part for #55 @psi29a  has created https://github.com/psi29a/ldaptor/tree/python3 but beside py3 changes it contains a lot of non-py3 changes.

That branch hasa conflict with latest master so make a future merge simpler this takes those non-py3 changes

You can see the diff here https://github.com/twisted/ldaptor/compare/master...psi29a:python3

Changes 
=======

I tried to only put non-controversial changes.

For some simple cases I tried to update the way raises is used, but this is not a complete change for all exception raising or catching.

In preparation for py3 unit testing I have renamed all assertEquals to assertEqual

How to check
=============

This is huge but I hope that most of the changes will only need time to check.

### DIFF
--- a/ldaptor/attributeset.py
+++ b/ldaptor/attributeset.py
@@ -1,14 +1,15 @@
 from copy import deepcopy
 
+
 class LDAPAttributeSet(set):
     def __init__(self, key, *a, **kw):
         self.key = key
         super(LDAPAttributeSet, self).__init__(*a, **kw)
 
     def __repr__(self):
-        values=list(self)
+        values = list(self)
         values.sort()
-        attributes=', '.join([repr(x) for x in values])
+        attributes = ', '.join([repr(x) for x in values])
         return '%s(%r, [%s])' % (
             self.__class__.__name__,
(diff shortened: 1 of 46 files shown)


psi29a:
LGTM, this needed to be done for readability. It's also a great exercise in reading and understanding the codebase. :)

psi29a:
Next PR, I would like to see an update to the Changelog if you could. Even for cleanups. :)

adiroiban:
> Next PR, I would like to see an update to the Changelog if you could. Even for cleanups. :)

I checked the CONTRIBUTING docs but there was no reference about docs/source/NEWS.rst

I will try to document it as part of the next PR

adiroiban:
> LGTM, this needed to be done for readability. It's also a great exercise in reading and understanding the codebase. :)

I haven't checked the codebase. For me this was mechanical work.
I agree that cleanup is needed... looking forward for adding pep8 checks to not regress on this.

PR #84, opened 2018-01-06
Update code to zope for py3 and not give errors on python3 compileall

adiroiban:
Scope
======

This is part of fixing #55

It builds on top of #83 and it does the minimum changes required to have the code compile on py3

Changes
=======

Do minimal changes to pass compileall

Update zope interface usage.

I have enabled the python3 tests to see what fails in preparation for future PRs.

As a drive-by I have re-arrange the order of some imports to have stdlib first, then other libs and last our own imports.

As a drive-by I fixed the indention of travis.yml to 2 spaces... it was a mix of 4 and 2 spaces.

How to check
=========

no errors should be present when compiling with python3

```
$ python3 -m compileall ldaptor
```

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -15,7 +15,7 @@ Changes
 
 - Using modern classmethod decorator instead of old-style method call.
 - Usage of zope.interfaces was updated in preparation for python3 port.
-- Code was updates to pass `python3 -m compileall` in preparation for py3 port.
+- Code was updated to pass `python3 -m compileall` in preparation for py3 port.
 
 Bugfixes
 ^^^^^^^^
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -14,6 +14,8 @@ Changes
 ^^^^^^^
 
 - Using modern classmethod decorator instead of old-style method call.
+- Usage of zope.interfaces was updated in preparation for python3 port.
+- Code was updates to pass `python3 -m compileall` in preparation for py3 port.
 
 Bugfixes
 ^^^^^^^^
--- a/ldaptor/samba/smbpassword.py
+++ b/ldaptor/samba/smbpassword.py
@@ -1,3 +1,4 @@
+import warnings
 try:
     import string
     maketrans = string.maketrans
--- a/ldaptor/samba/smbpassword.py
+++ b/ldaptor/samba/smbpassword.py
@@ -1,9 +1,14 @@
-import string, warnings
+try:
+    import string
+    maketrans = string.maketrans
+except ImportError:
+    maketrans = bytes.maketrans
+
 from ldaptor import md4, config
 
-lower = 'abcdefghijklmnopqrstuvwxyz'
+lower = b'abcdefghijklmnopqrstuvwxyz'
 upper = lower.upper()
-toupper=string.maketrans(lower, upper)
+toupper= maketrans(lower, upper)
 
 def nthash(password=''):
     """Generates nt md4 password hash for a given password."""
--- a/ldaptor/samba/smbpassword.py
+++ b/ldaptor/samba/smbpassword.py
@@ -1,16 +1,17 @@
 import string, warnings
 from ldaptor import md4, config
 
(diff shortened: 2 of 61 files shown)


psi29a:
Also LGTM, changelog entry needed to mark the updates to libraries please. :)

adiroiban:
Thanks for your review. That was fast :)

@psi29a ready for a new review as I have retargeted this to master... there were some unnecessary conflicts due to squash merging.

Let  me know if the level and wording for the relese notes are ok.

---

I will update the docs about contributing in a new branch where I also want to create a template for the PR.

ldaptor/samba/smbpassword.py
```diff
@@ -1,9 +1,15 @@
-import string, warnings
+import warnings
+try:
+    import string
+    maketrans = string.maketrans
+except ImportError:
```
adiroiban: While this does not gives errors, this is broken as here we should catch AttributeError

I am fixing it a branch which is a child of a child of this one so to simplify the merge  I think that is ok to merge it as it is... as with the squash merge things will get messy

ldaptor/test/test_attributeset.py
```diff
@@ -43,26 +42,31 @@ class TestSetOperations(unittest.TestCase):
     def testDifference(self):
         a = attributeset.LDAPAttributeSet('k', ['b', 'c', 'd'])
         b = attributeset.LDAPAttributeSet('k', ['b', 'c', 'e'])
-        self.assertEqual(a - b, sets.Set(['d']))
+        self.assertEqual(a - b, {'d'})
```
s0undt3ch: If ldaptor still supports 2.6, `set`s cannot be defined like this.
It's probably a good time to drop 2.6 support while gaining 3.x support?

docs/source/NEWS.rst
```diff
@@ -14,6 +14,8 @@ Changes
 ^^^^^^^
 
 - Using modern classmethod decorator instead of old-style method call.
+- Usage of zope.interfaces was updated in preparation for python3 port.
+- Code was updates to pass `python3 -m compileall` in preparation for py3 port.
```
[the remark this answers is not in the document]
adiroiban: thanks

ldaptor/test/test_attributeset.py
adiroiban: Thanks for the review.

I don't think that 2.6 is supported... and is hard to support both 2.6 and python 3

And 2.6 is no longer supported by upstream Python for a very long time

I see that tests are executed only under 2.7 https://github.com/twisted/ldaptor/commits/master/.travis.yml

I think 2.6 was already silently dropped :)

I see that in 14.0.0 it states that 2.6 is supported but it was removed without a changelog by @psi29a here https://github.com/twisted/ldaptor/commit/b6b4ea6fc55597977dbf5a64a6c8dca44eb12fe3#diff-354f30a63fb0907d4ad57269548329e3 more than 1 year ago

s0undt3ch: Then, awesome work!

psi29a: Twisted dropped 2.6 support, so we should follow their example.

psi29a:
Before I merge this in, there is a conflict to be resolved here:
.travis.yml

adiroiban:
I have fixed the merge in Travis. I can also merge it.

If we only want squash commit in merge, I think that we should disable all the other modes.

I have updated the settings of the repo. I have also removed wiki and projects as they are not used.  Let me know if you want the old settings back.

psi29a:
@adiroiban  Sure, I don't believe a wiki and the rest as necessary to this project since we have our own RTD.

psi29a:
We can just do merge, instead of squash and merge if that creates too many issues.  Just a fall-over from how I handle other projects where people will bring in a PR with many many commits which are just small fixes/typos.

adiroiban:
@psi29a I am ok with merge and squash. It will get extra work at merge time, but I prefer to have a clean history without "Fix typo." or "I am stupid" commit messages :) 

I merge it one but read the history many times :)

so +1 from me for the merge with squash.

PR #85, opened 2018-01-06
Update the dev docs to reflect the current process.

adiroiban:
Scope
=====

This updated dev docs to guide existing and new developers in the dev process.

Changes
======

Add GitHub PR template

To reduce the load on travis-ci, and the noise on the PR I am recomending to some tests on local machine to make sure the branch is green.

If you are working only on ldaptor you might remember that beside `tox -e pyflakes` you should also run the manifest checker, but for example I forgot to do that for this PR and only run pyflakes,

Update CONTRIB docs to talk about release notes requirements.

I have merged pyflakes and manifest checker into a single tox env. I find it easier to use.
Later we can add pep8 checks or other checks and we will not have to add new env each time and update docs to instruct devs to use them.

--------

The dev docs talk about code compatibility with Twisted style, but I think that is better to go with pep8.
It should require fewer changes ... and we will have a tool to check the style... the twisted codestyle checker is broken.

How to check
==========

I am still not sure what type of changes should go in release notes and which not.
If we want everything, maybe we should make sure PR merge messages are valid and auto-generate the release notes from that.

Check that the information is valid.

The github PR can't be check unless this is merged into master.

To see how MD looks you can use https://stackedit.io/editor

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,7 @@ env:
     - TOX_ENV=py27-tw160
     - TOX_ENV=py27-tw150
     - TOX_ENV=py27-tw140
-    - TOX_ENV=pyflakes
-    - TOX_ENV=manifest
+    - TOX_ENV=linters
 
 matrix:
   fast_finish: true
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,11 +1,26 @@
 How to Contribute
 =================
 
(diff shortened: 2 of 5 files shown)


psi29a:
This is a no brainer, merging. :)

Thank you for taking the time to update the less sexy side of things in ldaptor. It all needs love.

PR #86, opened 2018-01-06
Use absolute imports and fix all python2 -3 warnings when running tests.

adiroiban:
Scope
====

This update the code so that when running the tests with python2 -3 we don't get any deprecation warnings.

It also updates the usage of absolute imports.

This is a work toward #55

Changes
=======

The deprecation warning for running reactor.iterate() inside the test is still there.

It adds six as a dependecy to make it easier to port.
For now ldaptor.compat is still used. It can be removed in a final cleanup but for now I want to focus on the important parts. 

I think that six was already present in all ldaptor env as it might be a dependency of some of the existent direct dependencies.

------

with `op` and `foo2` and `a` and `b` I am using `sut` (System Under Test` for the variable holding the target of the test. For me is easier to identify what is tested.

I have also refactored some tests to use local setup/arrangement as it make it easier to read.

No need to switch between 2 methods to see how the tests works.

With local setup is also easier to make sure that setUp does not have unused code.... TestOperations was a good example of setUp code not used in any test.

------

While working at writing missing tests  I found some bugs, so I wrote more tests to make sure we are ok.

How to test
=========

I don't know how to update tox to pass arguments to python interpretor.

To test this I was doing.

```
. .tox/py27-twtrunk/bin/activate
python2 -3 .tox/py27-twtrunk/bin/trial ldaptor
```

and then check for warnings in ldaptor code.

This should be merged only after parent is merged and then re-targeted to be merged in master.

### DIFF
--- a/ldaptor/numberalloc.py
+++ b/ldaptor/numberalloc.py
@@ -1,4 +1,5 @@
 """Find an available uidNumber/gidNumber/other similar number."""
+from __future__ import division
 
 from ldaptor.protocols import pureldap
 
@@ -29,7 +30,7 @@ class freeNumberGuesser:
         if max is None:
             max=self.min+1000
 
-        guess=(max+self.min)/2
+        guess = (max + self.min) // 2
         d=self.makeAGuess(guess)
         d.addCallback(self._nextGuess, guess)
         return d
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+[run]
+branch = True
+source = ldaptor
+
+[paths]
+source =
+    ldaptor
+    */site-packages/ldaptor
+    *\site-packages\ldaptor
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 *.pyc
 *.pyo
 _trial_temp/
-.tox/
 ldaptor.test.*/
 ldaptor/test/ldif/webtests.tmp
 docs/*.png
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,14 @@ env:
   - TOX_ENV=py27-tw160
   - TOX_ENV=py27-tw150
   - TOX_ENV=py27-tw140
-  - TOX_ENV=py35-twtrunk
+  - TOX_ENV=py34-twtrunk
   - TOX_ENV=pyflakes
   - TOX_ENV=manifest
 
 matrix:
   fast_finish: true
   allow_failures:
-    - env: TOX_ENV=py35-twtrunk
+    - env: TOX_ENV=py34-twtrunk
 
 install:
   - pip install tox coveralls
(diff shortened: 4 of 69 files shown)


adiroiban:
should be ready for review. don't merge it as parent needs to be merged first.

this is another step toward #55

psi29a:
How can I see which is a parent and which is a child? It's isn't clear enough to me...

In your PRs can you link to other PRs and label them as parent or child so that we can follow the flow? 

If you already do this, my apologies... I didn't see anything indicating the parent/child relationship and merge order.

adiroiban:
@psi29a see 'adiroiban  wants to merge 2 commits into py3-compileall from py3-warnings'

This is where you see that "py3-warnings" is a child of "py3-compileall" and the merge order.

I will add links / mentions in the description

---------

@s0undt3ch I think that as long as we always run py2 and py3 tests there is no point in adding `absolute_import` ... if you accidentally put a  relative one a test should fail.

adiroiban:
I have retargeted it to master. Ready for review and merge :)

s0undt3ch:
:+1:

adiroiban:
I will take a look at this... 3 warnings were complaining about missing __hash__  but it looks like they are not used.

I will write some tests for the hash functions :)

adiroiban:
@s0undt3ch not sure what your thumb up mean... I am a bit old and I don't understand the meaning of all these emojis spread over the internet :)

Please be explicit if you are happy to see someone working on this, or that you have carefully reviewed the diff and you think that the PR is valid and can be merged.

BTW, my previous naive hash implementation was not correct.

Even if the tests were green, the changes had no tests.

I am working on fixing the object hash/equality and will submit this for review later.

This is not yet ready for review.

s0undt3ch:
@adiroiban first, I'm ecstatic about someone working on this, but my thumbs up was regarding the comment that absolute imports should not be necessary.
I made a mild review on the code, and, mild because I'm really not acquainted with the code base to say that this is good, it was more like, I love to read code, let's see the changes being made, I might learn a few more things along the way.

adiroiban:
@s0undt3ch No problem. I just wanted to make sure that I know what you mean :)

Meanwhile, with the power of some tools,  it should be easier to check that the porting is valid.

Now this should be ready for review.

I have not yet set up automated / continuous `python2 -3` warning testing, but this PR is already big and those test can be set up in a separate PR

ldaptor/numberalloc.py
```diff
@@ -29,7 +30,7 @@ def _nextGuess(self, found, lastGuess):
         if max is None:
             max=self.min+1000
 
-        guess=(max+self.min)/2
+        guess = (max + self.min) // 2
```
psi29a: why double (//) instead of single (/) ?

adiroiban: To keep the same behaviour between versions

```
Python 2.7.12 (default, Nov 20 2017, 18:23:56) 
>>> 10/3
3

Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
>>> 10/3
3.3333333333333335
>>> 10//3
3
```

psi29a: O.o  really?  These kinds of things drives me nuts...

PR #87, opened 2018-01-06
Fix module imports for py3 and use replacements for methods removed in py3.

adiroiban:
Scope
=====

This is another step toward #55 

This depends on #86

It tries to fix all the import errors and use replacements for things removed from py3

It only touches the simple things avoiding changes in the test part for py2.7

Changes
========

Use six for the moved code.

Implemented cmp() with (x > y) - (x < y). This PR is already big. In a separate PR I will address the complete replacement of the `__cmp__` functions as they are not used in py3

Some `__cmp__` were updated to meet coverage requirements see #93 and #94 

I have simplified the dns IP conversion.

I re-wrote some of the BERSequence tests in order to understand them and be able to write test for the change done in this PR.

It was importing a private stuff from twisted.mail.maildir import _generateMaildirName so I have replaced it with a UUID4. 

I have updated the way tox env are organized and updated Travis to always send coverage report... even on failures.

How to check
==========

Check that the changes are ok.

With this PR we should have the code compile on Py3 and the test run should not give import errors.

Besides the import changes, there are also some changes which are not mechanical and should be carefully reviewed.

There are still failing tests `FAILED (expectedFailures=1, failures=66, errors=172, successes=270)`
but from what I can see the remaining tests are due to Unicode vs bytes handling.

There are still about 20 other tests not related to str vs bytes but I will do them into a separate PR as they might require more work... and this PR is already big


ldaptor/config.py
```diff
@@ -32,7 +32,7 @@ def __init__(self,
             self.baseDN = baseDN
         self.serviceLocationOverrides = {}
         if serviceLocationOverrides is not None:
-            for k,v in serviceLocationOverrides.items():
+            for k, v in list(serviceLocationOverrides.items()):
```
s0undt3ch: If `serviceLocationOverrides` is not being modified while iterating, and ldaptor is already using `six`, use `six.iteritems(serviceLocationOverrides)` instead?

adiroiban: good point. Or we can leave this as it is in master  ... I was fooled by self.serviceLocationOverrides and thought that it is changed here.

will revert. thanks for looking into this :)

adiroiban:
I think that the coverage is ok.

Looks like Python3 coverage  is not reported... I am looking into this.

There is a flaky test which sometimes hits this line https://codecov.io/gh/twisted/ldaptor/compare/3e3e7cd209c350742f359e630721d6e1d3d9b560...8ec2aebbc971979729638a1f587b8e02a953fc5a/src/ldaptor/test/test_attributeset.py#L94

I am also checking the flaky test.

But meanwhile, this is ready for review... @psi29a do you have time to check this ? :)

s0undt3ch:
You removed the string case insensitive implementation... I'm curious with what you replaced it with...
I released https://github.com/saltstack/istr recently, that's py3 only though....

adiroiban:
> I'm curious with what you replaced it with

I have replaced it with more tests :) https://github.com/twisted/ldaptor/pull/87/files#diff-92fb3c6baa36fbba2b68feff0cf87b87R138

The default python implementation is already using case-insensitive matching so I don't know why we needed it.

From docs

> All option names are passed through the optionxform() method. Its default implementation converts option names to lower case.

The code has no history, but it was imported 14 years ago https://github.com/twisted/ldaptor/commit/c1d3935a5195787d4e89096ee23ab812963cc863#diff-8d40fa683b067b3721486a454252e52cR119

Let me know if can think of a test which fails with the current code.

thanks for the review

PR #88, opened 2018-01-07
Add codecov reporting for coverage.

adiroiban:
Scope
=====

This add codecov.io reporting when running on Travis

This  should give status for PR diff coverage.

Also, codecov.io can be configured via a file hosted on the repo so it should be much easier for other to configure it... no need to mess with webpage config.

Changes
=======

I set up the patch status to fail if diff is not 100%

I have also set up a custom report for the test code to make sure that we have 100%  test coverage... that is all tests are executed and we don't have tests which sit there and are not executed

I also kept the coveralls.

--------

As part of making the report useful and not have master red, I have updated the test code to have 100% coverage.

I have removed the .todo tests and move them to separate issues.
I find them annoying as after each run I see them fail... and then I have to ignore them. 

--------

A lot of the tests are synchronous and don't need to spin to reactor... so there is no point in returning a deferred and we can use the sync assertions like self.successResultOf and self.failureResultOf

How to test
=========

You should see the commit status reports from codecov.io

### DIFF
--- a/ldaptor/test/test_pureldap.py
+++ b/ldaptor/test/test_pureldap.py
@@ -617,12 +617,12 @@ class KnownValues(unittest.TestCase):
             result = klass(*args, **kwargs)
             result = str(result)
             result = map(ord, result)
-            if result!=encoded:
-                raise AssertionError(
-                      "Class %s(*%s, **%s) doesn't encode properly: "
-                      "%s != %s" % (klass.__name__,
-                                    repr(args), repr(kwargs),
-                                    repr(result), repr(encoded)))
+
+            message = (
+                "Class %s(*%r, **%r) doesn't encode properly: "
+                "%r != %r" % (
+                    klass.__name__, args, kwargs, result, encoded))
+            self.assertEqual(encoded, result, message)
 
     def testFromLDAP(self):
         """LDAPClass(encoded="...") should give known result with known input"""
@@ -1012,14 +1012,11 @@ class TestFilterSetEquality(unittest.TestCase):
         self.assertEqual(filter1, filter2)
 
     def test_escape_and_equal(self):
-        def custom_escaper(s):
-            return ''.join(bin(ord(c)) for c in s)
 
         filter1 = pureldap.LDAPFilter_and([
             pureldap.LDAPFilter_equalityMatch(
                 attributeDesc=pureldap.LDAPAttributeDescription('foo'),
                 assertionValue=pureldap.LDAPAttributeValue('1'),
-                escaper=custom_escaper
             ),
             pureldap.LDAPFilter_equalityMatch(
                 attributeDesc=pureldap.LDAPAttributeDescription('foo'),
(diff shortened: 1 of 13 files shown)


adiroiban:
not yet ready for review... I am still working on this

psi29a:
I see a report... ready for review? :)

adiroiban:
not ready.  It should be ready when all commit checks are green :)

I want to look at the test code and see if I can get 100% coverage.

There is no point in having tests just sitting there and which are never executed.

Right now I have trouble running the coverage report on my local tox environment...somehow the coverage is not collected.

adiroiban:
Also. if you look at codecov.io you can see that the pages are broken as it can't show the source...this is why I first wanted to get the reports on my local system

I see that tox.ini is set up to ignore the tests, but I have no idea why :)

Tests should be at least 1st class citizen code ... so have the coverage reported just like any other code :)

I got the reporting working outside of tox so now I am working at cleaning the tests so that we don't have code which is not executed

psi29a:
If it helps, have a look at how we use tox in additional to travis in the TxMongo project:
https://github.com/twisted/txmongo/blob/master/.travis.yml

For example, we explicitly set the ENV when setting up tox which includes telling travis to use not just python 2.7, but other python versions as well. Perhaps going this route will help things?

adiroiban:
While working on this I have accidentally pushed something to master (I have soon reverted it)

------

I think that we should use the nice GitHub feature and protect master.

I have enabled them. Let me know what do you think and I can revert the changes.

-------

Do we need both PR and PUSH tests for Travis?


I will keep coveralls for now, for comparison, but I guess that once we are happy with codecov.io we can remove it.

-------

I made some progress and all tests have 100% coverage.

I am confused about ldaptor.testutil vs ldaptor.test.util . I have created #91 to look at this

---------

This should be ready for review.

adiroiban:
@psi29a this is ready for review.

Do you get coverage reporting on your local system when running `tox -e py27-twtrunk` ?

ldaptor/test/test_pureldap.py
```diff
@@ -617,12 +617,12 @@ def testToLDAP(self):
...
-                      "Class %s(*%s, **%s) doesn't encode properly: "
-                      "%s != %s" % (klass.__name__,
-                                    repr(args), repr(kwargs),
-                                    repr(result), repr(encoded)))
+
+            message = (
+                "Class %s(*%r, **%r) doesn't encode properly: "
+                "%r != %r" % (
+                    klass.__name__, args, kwargs, result, encoded))
+            self.assertEqual(encoded, result, message)
```
psi29a: should we eventually migrate to the string.format() way instead of the printf style?

adiroiban: As a drive by, we can use string.format()

Careful with string format as you should not pass unsanitized input string as the template.
See http://lucumr.pocoo.org/2016/12/29/careful-with-str-format/

So for simple things, I still use % out of reflex

adiroiban:
Thanks for the review

Well, I need to have good coverage report to help when doing the py3 porting.
I am doing test-driven porting, so is critical that we have tests for the code which is touched.

ldaptor/test/test_pureldap.py
psi29a: Sanitizing user input should happen before it is used in code anyway.

I had a read through the article, but it isn't anything I wasn't aware of. So long as we are in charge of the code, the objects that are being created by user input, which we are, and we're not doing anything weird with the object or pod when passing it into the string.format() then I'm not worried.

string.format (and other print functions) get a bad rap because they are the things that expose the problem to the end-user or developer anyway. ;)

PR #92, opened 2018-01-11
Fix coverage reporting for local dev environment.

adiroiban:
Scope
=====

This updates the development environment to report coverage.

Based on that it will produce coverage reports in HTML, as well as a diff coverage since master.

Changes
========

The important thing was to run trial with '-m' and in this way the coverage was reported.
I think that coverage was confusing ldaptor source with ldaptor package.

I have created a dedicated configurate for coverage so that we don't have to pass a long list of arguments.

I have moved the whole .tox files to build/ as this should be the standard folder for dev related things.

On the build folder I am putting the HTML reports, and all other intermediary files.

I have recognized the tox file so that the linters are also run for the dev environment.
In this way it should be easier to detect errors and reduce the change of pushing bad commits, and creating extra work for Travis-CI

While working on this, I have tested this by working on ldaptor/attributeset.py

I have removed the unused methods as in py2.7 we have different methods.
It looks like the test pass with default set inheritance.

How to test
========

Check that the updated dev docs make sense and that you get coverage reporting on local tox, as well as in buildbot

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+[run]
+branch = True
+source = ldaptor
+
+[paths]
+source =
+    ldaptor
+    */ldaptor
+    *\ldaptor
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 *.pyc
 *.pyo
 _trial_temp/
-.tox/
 ldaptor.test.*/
 ldaptor/test/ldif/webtests.tmp
 docs/*.png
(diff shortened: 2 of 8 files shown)


psi29a:
Should we give those PRs with -0.01% a pass?

adiroiban:
> Should we give those PRs with -0.01% change a pass?

Not before checking what is going on there :)

Somehow not all changes were pushed to GitHub... I have pushed them now.

------

I think we should only enable travis-ci/push for commits in master as otherwise we have 2 similar builds for each PR

@psi29a let me know what do you think

adiroiban:
This is now ready for review.

Maybe we should start using labels to mark the state of a PR

PR #96, opened 2018-01-16
Fix non str vs bytes tests for Python3 porting.

adiroiban:
Scope
=====

Another one for #55 

This should be the last opportunistic porting to Python3.
It tried to handle all the remaining test failures which are not due to str vs bytes differences between py2 and py3

For now, I am focusing the porting for python3.5.. once 3.5 is done we can look at other versions.

Changes
=======

I have added a new travis-ci env to run tests on py3.5 and make sure we don't add regressions.

Some tests have explicit bytes markup for strings but they still fail... they will be fixed in follow up PR.

I did some drive by fixes for marking bytes while reading the code and working on fixing other tests.
As long as coverage is 100% i hope that this is ok.
I don't have much time, and I prefer to do the changes now, rather than revert them and re-read the code and check it later.

This does a quick porting of distinguishedname, but after this is merged I plan to have a dedicated PR for doing a proper porting of distinguishedname . Once this is done more tests will pass.

With this PR we have 400 passing tests in comparison with  306 which are in master or a total of 522 tests

For comparison with unknown types, I went for Py3 behaviour and raise TypeError.

Other infra changes:

* coveralls was removed
* A new py35-ported tox env was added to help preventing py3 regresssion for the code which was already green

How to test
=========

Check that changes make sense.

Check the new tests, as they might not assert the right thing.
Some code did not had docstring so I wrote the test based on my understanding of how I think that it should behave.


ldaptor/protocols/pureber.py
```diff
@@ -210,10 +209,16 @@ def __init__(self, value=None, tag=None):
         self.value = value
 
     def __str__(self):
-        value = str(self.value)
-        return chr(self.identification()) \
-               + int2berlen(len(value)) \
-               + value
+        if six.PY3 and isinstance(self.value, bytes):
```
s0undt3ch: We should probably default to checking for py2 instead of py3, because eventually we'll have py4...

psi29a: Agreed.

psi29a:
Would love to merge, but as said above... we should be checking on PY2 since this is the one to be deprecated and eventually removed.

.travis.yml
```diff
@@ -4,31 +4,32 @@ language: python
...
+    env: TOX_ENV=py27-test-twtrunk
+  - python: 3.5
+    env: TOX_ENV=py35-ported
+  - python: 3.5
+    env: TOX_ENV=py35-test-twtrunk
+  - python: 2.7
+    env: TOX_ENV=linters
   allow_failures:
-    - env: TOX_ENV=py34-test-twtrunk
+    - env: TOX_ENV=py35-test-twtrunk
```
psi29a: py36 instead of py35?

adiroiban:
> Would love to merge, but as said above... we should be checking on PY2 since this is the one to be deprecated and eventually removed.

This is not yet ready.
Before merging this I think that we should have 100% coverage for the patch as a minimum check for regressions.

And yes, it should be updated to python3.6.

But I don't know when I will have time to work on this.

.travis.yml
adiroiban: Well, in master we have tox.ini with `{py33,py34,py35}-test-{twtrunk,dev}`
This is why I focused on py35.

The main part here is `TOX_ENV=py3N-ported` which is a step toward making sure we don't regress on the tests that were already ported

adiroiban:
and there is still one more tests which can be fixed here as it is not related to str vs bytes

```
builtins.AttributeError: 'getset_descriptor' object has no attribute 'tb_frame'

ldaptor.test.test_ldiftree.Dir2LDIF.testEmptyError
```

adiroiban:
This is now ready for review :)

A big PR as I had to write new tests while doing the code updates for Python3.

In this branch we have `FAILED (failures=72, errors=102, successes=447)`

s0undt3ch:
So awesome!

psi29a:
@cwaldbieser we have our work cut out for us ;)

cwaldbieser:
I've made it from the beginning of the changes through `ldaptor/test/test_pureber.py` so far.  I'm not sure if there is some way integrated into GitHub to annotate my questions, so I will put them in the comments section for now.  Mostly, I just have questions about some changes that were made when they don't suggest an obvious reason why the change was made.  Lots of the changes I saw actually did make a lot of sense and made the code a lot easier to read.

---------------------------------------------
docs/source/examples/ldaptor_with_upn_bind.py
---------------------------------------------

* (line 15) Why escape the backslashes in the example LDIF text?

----------------
ldaptor/entry.py
----------------

* (line 22) Why is the salt changed from 8 bits to 7 bits?

--------------------------------------
ldaptor/test/test_distinguishedname.py
--------------------------------------

* (lines 158-159 original) In class class `LDAPDistinguishedName_RFC2253_ExamplesBytes` - The original code seemed to suggest this could be initialized with Unicode characters outside the ASCII set, but the new code seems to suggest this is not the case.  Why?

-----------------------------
ldaptor/test/test_ldiftree.py
-----------------------------

* In test `test_subtree_many_cb()`, I noticed the original collected the results, sorted, and compared.  Your test just counts to make sure the same number of components are in the result.  This isn't exactly the same test.  Does another test check to make sure that the components that are returned are the correct components?

----------------------------
ldaptor/test/test_pureber.py
----------------------------

* (line 36) - Is there some reason for `return [six.byte2int([x]) for x in s]` instead of `return [six.byte2int(x) for x in s]`?  I'm not sure why the extra list ...

-----
Misc
-----

What does the symbol `sut` stand for in the tests.  I see `o` was replaced multiple times with `sut`, but I'm not sure what it means ("some uvver ting"?).

Still a lot of code to go through yet.  It is a rather big PR.

adiroiban:
Thanks for looking into this and sorry for the big PR

You should be able to leave comments inline. I will try to respond to some questions inline.

> What does the symbol sut

sut is `system under test` ... I used it as a convention to make it easier what is the target of the test.
sometimes the tests have a lot of objects in the arrange part...and then the act part involved the interaction of multiple objects and then in the assert you might check some mock or test double and then you have to read and re-read a test to see what is the target of that test.

docs/source/examples/ldaptor_with_upn_bind.py
```diff
@@ -12,7 +12,7 @@
 gn: Bob
 sn: Roberts
 mail: bob@example.org
-homeDirectory: e:\Users\bob
+homeDirectory: e:\\Users\\bob
```
adiroiban: They need to be escaped as the  docstring is not marked as row and then python tried o resolve some `\UXXXX` Unicode value :)

ldaptor/entry.py
```diff
@@ -16,16 +16,20 @@
...
 
 def sshaDigest(passphrase, salt=None):
+    """
+    Return the salted SHA for `passphrase` which is passed as bytes.
+    """
     if salt is None:
         salt = ''
         for i in range(8):
-            salt += chr(random.randint(0, 255))
+            salt += chr(random.randint(0, 127))
```
adiroiban: this was changed to generate ascii characters... and make it easy to convert later to bytes.
I hope nobody is using this for production and this here just for demonstration purpose... i didn't wanted to spend more time with this

ldaptor/test/test_distinguishedname.py
```diff
@@ -154,13 +160,25 @@ class LDAPDistinguishedName_RFC2253_Examples(TestCaseWithKnownValues):
          [[('1.3.6.1.4.1.1466.0', '#04024869')],
           [('O', 'Test')],
           [('C', 'GB')]]),
+        )
 
-        (u'SN=Lu\u010di\u0107'.encode('utf-8'),
-         [[('SN', u'Lu\u010di\u0107'.encode('utf-8'))]])
```
adiroiban: This is still possible, but it was moved into a dedicated tescase with a dedicated docstring see LDAPDistinguishedName_UTF8_Init

ldaptor/test/test_ldiftree.py
```diff
@@ -572,9 +570,7 @@ def _cb_test_subtree_many_cb(self, r, got):
             self.bar,
             self.foo,
             ]
-        got.sort()
-        want.sort()
-        self.assertEqual(got, want)
+        six.assertCountEqual(self, expected, got)
```
adiroiban: > In test test_subtree_many_cb(), I noticed the original collected the results, sorted, and compared. Your test just counts to make sure the same number of components are in the result. This isn't exactly the same test. Does another test check to make sure that the components that are returned are the correct components?

I think that `assertCountEqual` does the same thing.
It has a stupid name, but see docs https://docs.python.org/3.2/library/unittest.html#unittest.TestCase.assertCountEqual

ldaptor/test/test_pureber.py
```diff
@@ -30,9 +30,11 @@ def s(*l):
         r = r + str(e)
     return r
 
+
 def l(s):
     """Split a string to ord's of chars."""
-    return map(lambda x: ord(x), s)
+    return [six.byte2int([x]) for x in s]
```
adiroiban: > (line 36) - Is there some reason for return [six.byte2int([x]) for x in s] instead of return [six.byte2int(x) for x in s]? I'm not sure why the extra list ...

I think that byte2int requires a list https://pythonhosted.org/six/#six.byte2int as it needs a byte string as input

adiroiban:
@cwaldbieser at https://github.com/twisted/ldaptor/pull/96/files you can move the mouse next the the line number (in either sides) and it will add a coment on that line... when that line is modified by a future commit the comment will be considered obsolete and no longer visible... something like "resolved"

You can group multiple comments in a review. and then you can click the "Review changes" button to leave a final comment and status like approved or need more work.

Thanks!

ldaptor/test/test_pureber.py
cwaldbieser: I'm pretty certain `bs` is just a byte string, though a list with a single byte char seems to work:

	waldbiec@waldbiec-Latitude-E5570:~/git-repos/ldaptor$ /usr/bin/python2 -V
	Python 2.7.12
	waldbiec@waldbiec-Latitude-E5570:~/git-repos/ldaptor$ /usr/bin/python3 -V
	Python 3.5.2
	waldbiec@waldbiec-Latitude-E5570:~/git-repos/ldaptor$ cat test.py
	#! /usr/bin/env python

	import six


	def main():
	    chars = b'chars'
	    l = [chars[0]]
	    print(six.byte2int(chars))
	    print(six.byte2int(l))

	if __name__ == "__main__":
	    main()
	waldbiec@waldbiec-Latitude-E5570:~/git-repos/ldaptor$ /usr/bin/python2 test.py 
	99
	99
	waldbiec@waldbiec-Latitude-E5570:~/git-repos/ldaptor$ /usr/bin/python3 test.py 
	99
	99

Unless there's something subtle here that I'm missing.

ldaptor/samba/smbpassword.py
```diff
@@ -68,7 +72,7 @@ def _deshash(p):
     # Insert parity bits. I'm not going to bother myself with smart
     # implementations.
     bits = []
-    for byte in [ord(c) for c in p]:
+    for byte in [six.byte2int([c]) for c in p]:
```
cwaldbieser: My `six.byte2int()` nitpick, again.

ldaptor/test/test_ldiftree.py
cwaldbieser: Yes-- you are correct.  I like the Py2 name better (`assertItemsEqual()`).

ldaptor/test/test_pureldap.py
```diff
@@ -28,14 +26,16 @@ def s(*l):
...
+        if isinstance(e, int):
+            e = chr(e)
+        r = r + str(e)
     return r
 
+
 def l(s):
     """Split a string to ord's of chars."""
-    return map(lambda x: ord(x), s)
+    return [six.byte2int([x]) for x in s]
```
cwaldbieser: Another `six.byte2int()` invocation taking a list when the plain old byte string should work.  I am pretty sure I saw at least one place in the code where this function was being used with a plain old byte string.

ldaptor/test/test_pureldap.py
```diff
@@ -619,7 +619,7 @@ def testToLDAP(self):
         for klass, args, kwargs, decoder, encoded in self.knownValues:
             result = klass(*args, **kwargs)
             result = str(result)
-            result = map(ord, result)
+            result = [ord(x) for x in result]
```
cwaldbieser: Just wondering why sticking with `ord()` here and not `six.byte2int()`.

psi29a:
It looks good to me, waiting on merge to see if there will be another round to address @cwaldbieser 's comments.

ldaptor/test/test_pureldap.py
adiroiban: The test seems to pass...and for this PR this is what I was aiming :)

ldaptor/test/test_pureldap.py
adiroiban: I did a search, but I did not found any six.byte2int used without a list.

adiroiban:
Thanks, @cwaldbieser for the review.

Note that I am doing the Python 3 porting for fun.
I am only using Python2 for production and have no plan to use python3.
So there are many things which I am not doing right...but might work with the tests.


-----------

In the  tests, `six.byte2int` does not receive a direct bytestring and in Python 3 we have

```
Python 3
>>> [a for a in b'boo']
[98, 111, 111]]

Python 2.7.12 (default, Dec  4 2017, 14:50:18) 
>>> [a for a in b'boo']
['b', 'o', 'o']

```

```

Note that I got this error

```
ldaptor/test/test_pureber.py", line 36, in <listcomp>
    return [six.byte2int(x) for x in s]
builtins.TypeError: 'int' object is not subscriptable
```

This is from 

```
    knownValues=(
        ("", [0x04, 0]),
        ("foo", [0x04, 3]+l(b"foo")),
        (100 * "x", [0x04, 100]+l(100 * b"x")),
        (ObjectThatCanStringify(), [0x04, 3]+l(b"bar")),
        )
```

and similar error on `ldaptor.test.test_smbpassword.TestLMHash.testKnownValues`


----------

adiroiban:
> It looks good to me, waiting on merge to see if there will be another round to address @cwaldbieser 's comments.

@psi29a Hi. On what merge are you waiting? I am confused :)

I don't plan to make more changes on this PR

Thanks!

psi29a:
Sorry, I meant this particular PR. Waiting on your reply to @cwaldbieser 's comments

adiroiban:
> Sorry, I meant this particular PR. Waiting on your reply to @cwaldbieser 's comments

@psi29a I have replied to all the comments.

What did a miss? :)

psi29a:
I should learn to press F5 more often... ;)

Merging! Thanks!

adiroiban:
> I should learn to press F5 more often... ;)

I think that I have commented after you... and somehow the messages were out of order on my side... so F5 was also required on my side

PR #103, opened 2018-06-05
Paged search

cwaldbieser:
I updated the `LDAPClient` API by adding `send_multiResponse_ex()` which allows you to send LDAP controls along with your request.  The new method also makes sure that any LDAP response controls can be sent back to your client handlers.  This is specifically useful for implementing paged search requests.  Paged search requests are particularly important if you are making searches against an Active Directory DIT and expect a lot of search results.  By default, AD limits you to around ~1000.  However, if your client uses pages searches, it can retrieve all the results.

I didn't want to simply change the existing `send_multiResponse()`, as I didn't want to unnecessarily break any client code.

I also ended up updating `ldapsyntax.LDAPClient` so that it's `search()` method can take advantage of this new capability, because constructing your own `pureldap.LDAPSearchRequest` instance from scratch is a bit much.

I also added a new recipe that shows how the new capability can be used.
### Contributor Checklist:

* [x ] I have updated the release notes at `docs/source/NEWS.rst` 
* [x ] I have updated the automated tests.
* [ x] All tests pass on your local dev environment. See `CONTRIBUTING.rst`.


psi29a:
linter not happy, codecov also seems unhappy

cwaldbieser:
I had been running `pycodestyle` (formerly `pep8`) on the changed files and making lots of little style updates.  This unfortunately affected coverage.  I think I've got the hang of using `tox`, now.  I reverted some of the style changes.

Looks like I'm closer-- 99.31%.  I'll get it eventually ...

cwaldbieser:
I'm wondering if this had something to do with multiple commits confusing codecov?  I'm going to try squashing them into a single commit without so much churn and submit that as a new PR.  If that works, we can just reject this one.

cwaldbieser:
The new PR is #104 .  I guess I'll find out soon if the coverage report is any better.  At least the commit history is easier to read.

PR #106, opened 2018-06-18
Fix proxy race condition with immediately disconnecting clients (#105)

### LINKED ISSUE #105: Proxy may keep connection to proxied server open if client disconnects immediately
Hello! We're developing an LDAP proxy application based on ldaptor which has worked pretty well so far, so thanks for the great library! We recently stumbled upon a small bug which can be reproduced as follows:

* Use the current ldaptor master (7a51037ccf3ebe387766a3a91cf9ae79a3a3393d)
* Run the exemplary logging proxy code from [the docs](https://ldaptor.readthedocs.io/en/latest/cookbook/ldap-proxy.html), proxying any LDAP server (AD at ``10.0.1.161:7389`` in our case)
* Use netcat to perform a simple "health check": ``netcat -z localhost 10389``. The ``-z`` flag is [described](https://www.irongeek.com/i.php?page=backtrack-3-man/netcat) as follows:  _In connect mode it means that as soon as the port is open it is immediately shutdown and closed._
* Even though netcat has disconnected, ``netstat -nto`` reports a connection from the LDAP proxy to the proxied server in the ESTABLISHED state:
```
tcp        0      0 10.1.0.4:60478          10.0.1.161:7389         ESTABLISHED off (0.00/0/0)
```
This socket is only closed when the LDAP proxy is shut down. This is unfortunate because some appliances may perform periodical "health checks" against the LDAP proxy, similar to the netcat call above ([see here](https://community.privacyidea.org/t/too-many-established-tcp-connections-from-the-ldap-proxy/718)). If every such health check persists one connection to the proxied server, the LDAP proxy will eventually run out of open files. 

I believe this is due to a race condition:
* When netcat connects to the proxy, ``connectionMade`` is called, which initiates a connection to the proxied server. It uses a deferred to invoke ``_connectedToProxiedServer`` once the connection has been established: https://github.com/twisted/ldaptor/blob/7a51037ccf3ebe387766a3a91cf9ae79a3a3393d/ldaptor/protocols/ldap/proxybase.py#L40
* netcat closes the connection to the proxy, so ``connectionLost`` is called. However, as netcat does so *immediately*, the above deferred's callback has not been called yet, so ``self.client`` is still None.
* After that, ``_connectedToProxiedServer`` is finally called. Because of that, the connection to the proxied server is never closed.

I'll come up with a small PR to fix this, though it may not be the optimal solution as I'm not that familiar with the ldaptor codebase :)

fredreichbier:
This should fix #105.

I'm not sure if this completely rules out the race condition: This only handles clients that disconnect without having sent any data. Might a client be able to even send some LDAP requests before the connection to the proxied server has been established? In that case, they are added to ``queuedRequests``, and the connection persists.

I'm happy about any comments :)

### Contributor Checklist:

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -32,6 +32,7 @@ Bugfixes
 ^^^^^^^^
 
 - DN matching is now case insensitive.
+- Proxies now terminate the connection to the proxied server in case a client immediately closes the connection.
 
 
 Release 16.0 (2016-06-07)
--- a/ldaptor/protocols/ldap/proxybase.py
+++ b/ldaptor/protocols/ldap/proxybase.py
@@ -62,10 +62,11 @@ class ProxyBase(ldapserver.BaseLDAPServer):
             return d
         else:
             self.client = proto
-            if not self.connected and not self.queuedRequests:
-                # Client has disconnected in the meantime without sending any requests
+            if not self.connected:
+                # Client no longer connected, proxy shouldn't be either
                 self.client.transport.loseConnection()
                 self.client = None
+                self.queuedRequests = []
             else:
                 self._processBacklog()
 
--- a/ldaptor/protocols/ldap/proxybase.py
+++ b/ldaptor/protocols/ldap/proxybase.py
@@ -62,7 +62,12 @@ class ProxyBase(ldapserver.BaseLDAPServer):
             return d
         else:
             self.client = proto
-            self._processBacklog()
+            if not self.connected and not self.queuedRequests:
+                # Client has disconnected in the meantime without sending any requests
+                self.client.transport.loseConnection()
+                self.client = None
+            else:
+                self._processBacklog()
 
     def _establishedTLS(self, proto):
         """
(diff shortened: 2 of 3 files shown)


cwaldbieser:
I'm thinking that if the connection is closed by the client, any pending requests it may have sent should be discarded.  Having a pending request shouldn't stop the proxy from closing the connection.  So I'd suggest changing your correction slightly:

            if not self.connected:
                # Client is no longer connected.  Proxy shouldn't be either!
                self.client.transport.loseConnection()
                self.client = None
		self.queuedRequests = []
            else:
                self._processBacklog()

Thanks for catching that!

fredreichbier:
Hi Carl, thanks for having a look and your comment! Discarding all queued requests makes sense, so changed the code accordingly. :)

cwaldbieser:
Looks good to me.  I ran a functional test using the ldaptor server example and the proxy example, and it still behaved the way I expect.  I'll merge the PR.

PR #114, opened 2018-10-08
Py3compatibility

GrayAn:
Greetings! I made an attempt in porting some of the library code to the Python 3. It may be helpful in finding the better approach in this str/unicode/bytes problem. The key concepts I use:

1. `__str__` methods in pureber and pureldap modules were renamed to `__bytes__`. `__str__` is now used as an alias for `__bytes__` method in the Python 2. This modules allowed implicit conversion to the Python 2 str if unicode was used for their classes initializing so I tried to preserve this feature though there are no unicode tests for these modules.
2. distinguishedname.py classes allow using of both str/bytes and unicode/str for initialization. Their values are stored in unicode/str but both `__bytes__` and `__str__` methods are used.
3. `str()` conversion was changed to `bytes()` in many modules and tests.

The main problem I have stumbled is encoding/decoding of LDAP attributes and values. If we add an attribute to an entry do we need to convert it to str/bytes or unicode/str? If we parse a received entry what type should we use for its attributes and values? This ambiguity prevented me from making proper fixes for LDIF modules and most importantly ldapserver.

No tests were broken for the Python 2.7. The current test picture for the Python 3.5 and 3.6 is:

```
ldaptor/test/test_attributeset.py............
ldaptor/test/test_autofill.py.
ldaptor/test/test_autofill_posix.py..
ldaptor/test/test_autofill_samba.py............
ldaptor/test/test_config.py............
ldaptor/test/test_connector.py...
ldaptor/test/test_delta.py.........................................
ldaptor/test/test_distinguishedname.py.......................
ldaptor/test/test_dns.py...............
ldaptor/test/test_entry.py................
ldaptor/test/test_examples.pyFFF
ldaptor/test/test_fetchschema.py.
ldaptor/test/test_inmemory.py.......................................
ldaptor/test/test_ldapclient.py...............
ldaptor/test/test_ldapfilter.py...........................................................
ldaptor/test/test_ldapsyntax.py................................................................
ldaptor/test/test_ldifdelta.pyF.....F.....FFFF.
ldaptor/test/test_ldifprotocol.py...FFF.FFFF..F..F
ldaptor/test/test_ldiftree.py.....FFFFF.F.....................F.......FFFF.FF.....
ldaptor/test/test_match.py...............................
ldaptor/test/test_merger.py........
ldaptor/test/test_proxy.py..F.
ldaptor/test/test_proxybase.py.......F.
ldaptor/test/test_pureber.py...............................
ldaptor/test/test_pureldap.py.....................
ldaptor/test/test_schema.py.......
ldaptor/test/test_server.py.FF.......F..F..FF.FFFFFF..FFFF.x
ldaptor/test/test_smbpassword.py...
ldaptor/test/test_svcbindproxy.py......
ldaptor/test/test_usage.py.....

49 failed, 513 passed, 1 xfailed in 21.86 seconds
```

Contributor Checklist:

I have not updated the release notes as these changes look like already mentioned "preparation for py3 port" in NEWS.rst .

### DIFF
--- a/ldaptor/protocols/ldap/distinguishedname.py
+++ b/ldaptor/protocols/ldap/distinguishedname.py
@@ -11,14 +11,21 @@ escapedChars_leading = r' #'
 escapedChars_trailing = r' #'
 
 
+def _to_bytes(s):
+    if isinstance(s, six.text_type):
+        return s.encode('utf-8')
+    return s
+
+
+def _to_unicode(s):
+    if isinstance(s, six.binary_type):
+        return s.decode('utf-8')
+    return s
+
+
 def escape(s):
     r = ''
     r_trailer = ''
-
-    have_bytes = False
-    if isinstance(s, bytes):
-        s = s.decode('utf-8')
-        have_bytes = True
 
     if s and s[0] in escapedChars_leading:
         r = '\\' + s[0]
@@ -36,71 +43,50 @@ def escape(s):
         else:
             r = r + c
 
-    result = r + r_trailer
-    if have_bytes:
-        return result.encode('utf-8')
-    else:
-        return result
+    return r + r_trailer
 
 
 def unescape(s):
-    r = ''
-
-    have_bytes = False
-    if isinstance(s, bytes):
-        s = s.decode('utf-8')
-        have_bytes = True
+    r = bytearray()
 
     while s:
-        if s[0] == '\\':
-            if s[1] in '0123456789abcdef':
-                r = r + chr(int(s[1:3], 16))
+        if s[0:1] == b'\\':
+            if s[1] in b'0123456789abcdef':
+                r.append(int(s[1:3], 16))
                 s = s[3:]
             else:
-                r = r + s[1]
+                r.append(s[1])
                 s = s[2:]
         else:
-            r = r + s[0]
+            r.append(s[0])
             s = s[1:]
 
-    if have_bytes:
-        return r.encode('utf-8')
-    else:
-        return r
+    return bytes(r)
 
 
 def _splitOnNotEscaped(s, separator):
     if not s:
         return []
 
-    have_bytes = False
-    if isinstance(s, bytes):
-        s = s.decode('utf-8')
-        have_bytes  = True
-
-
-    r = ['']
+    r = [b'']
     while s:
         first = s[0:1]
 
-        if first == '\\':
+        if first == b'\\':
             r[-1] = r[-1] + s[:2]
             s = s[2:]
         else:
 
             if first == separator:
-                r.append('')
+                r.append(b'')
                 s = s[1:]
-                while s[0:1] == ' ':
+                while s[0:1] == b' ':
                     s = s[1:]
             else:
                 r[-1] = r[-1] + first
                 s = s[1:]
 
-    if have_bytes:
-        return [x.encode('utf-8') for x in r]
-    else:
-        return r
+    return r
 
 
 class InvalidRelativeDistinguishedName(Exception):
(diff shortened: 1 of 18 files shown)


adiroiban:
Thanks @GrayAn 

What is the scope of this PR? Maybe you can update the title of this PR to make it clear what is the target of this PR.

Note that there is py35-ported env in tox.ini https://github.com/twisted/ldaptor/blob/8f150c32d82ae58c5492dda1b1b50f011b69c15a/tox.ini#L55 which contains the code ported so far.

It would help to update that list with your work to make sure there will be no py3 regression on this.

Please check the codecov coverage report. I see that some code has not coverage.
Not sure if there are missing tests or is just code which is not used and can be removed.

-----------

To make the review process easier and to make sure we don't introduce any regression is best to create small PR dedicated on porting a specific API/module

--------

I remember working on the Py3 porting for ldaptor and the next step on my porting path was to sort the LDAPattribute LDAPValue API.

So before porting anything else, we might want to focus on getting that API ported in a separate PR and in that PR we can see how to port them.

---------

Regarding bytes vs Unicode.

In an ideal world:

Anything that is read / written to ldapton on/from wire/disk/io is byte.
Input/output via library's API is Unicode / base64 (for bytes)
Once something was read by ldaptor as bytes, it is converted to unicode and handled internally as unicode.

Now, the current py2 implementation reads/writes bytes and does internal handling as bytes
Also, all API calls are bytes.
The work for python3 is updating the internal implementation to work with Unicode and update the API to access Unicode and output Unicode.

----------

To make porting easier, we can also go with bytes everywhere and then code calling/reading API will need to make its own conversion.

Maybe @psi29a can weight in on the porting path.

adiroiban:
> Don't know how to fix code coverage as almost all of missing lines is the Python 3 code for example:

I see 2 options to fix it:

1. write new tests for that branch .... new py3 tests
2. if writing test does not make sense, remove that code as it means that the code is not needed

---------

But, can we not have Py3/py2 specific code execution branches?

I don't like that `__str__` is used in ldaptor for getting the wire representation.

My understanding is that `__str__` is for getting a friendly name/represenation of the object, and not its wire serialization. https://docs.python.org/3/reference/datamodel.html#object.__str__

So, I would say that we should blow `__str__` and make sure it will never return the wire representation.

I think that we should use `__bytes__` everywhere, including on Python2.
It is more explicit and it matches the documentation.

I am fine with `__bytes__` but for my taste, I would prefer something even more explicit like `toWire` :) 

-------

The title is still very generic for me and does not help describe the scope of this PR :)

But, since there are no regression for Python2.7, as long as the coverage is 100% , I think that we can review and merge it as it is now.

ldaptor/protocols/ldap/distinguishedname.py
```diff
@@ -11,15 +11,22 @@
 escapedChars_trailing = r' #'
 
 
+def _to_bytes(s):
```
adiroiban: this looks like duplicate code...and is generic/utility code...so I think that it can be moved outside of this module.

not sure why six does not have yet this helper :)

ldaptor/protocols/ldap/distinguishedname.py
```diff
@@ -124,23 +110,27 @@ def __init__(self, stringValue=None, attributeType=None, value=None):
...
-            if isinstance(stringValue, bytes):
-                equal_char = b'='
-            else:
-                equal_char = '='
+            stringValue = _to_unicode(stringValue)
 
-            if equal_char not in stringValue:
+            if '=' not in stringValue:
+                if six.PY2:
+                    stringValue = _to_bytes(stringValue)
```
adiroiban: why do we use `_to_bytes`
Once ported, the code should always handle internal data as Unicode.

That is in Py2 and Py3 we will always handle internal data as Unicode.

ldaptor/protocols/ldap/distinguishedname.py
```diff
@@ -124,23 +110,27 @@ def __init__(self, stringValue=None, attributeType=None, value=None):
...
                 raise InvalidRelativeDistinguishedName(stringValue)
-            self.attributeType, self.value = stringValue.split(equal_char, 1)
+            self.attributeType, self.value = stringValue.split('=', 1)
 
     def __str__(self):
-        return '='.join((escape(self.attributeType), escape(self.value)))
+        if not six.PY2:
+            return self.__bytes__().decode('utf-8')
+        return self.__bytes__()
```
adiroiban: I think that we should not return bytes here, but a real human readable representation of this object.
In this way, we can make sure that any usage of `__str__` with the intended purpose of getting the wire representation will fail

ldaptor/protocols/ldap/distinguishedname.py
```diff
@@ -192,25 +182,30 @@ def __init__(self, magic=None, stringValue=None, attributeTypesAndValues=None):
             assert attributeTypesAndValues is None
             if isinstance(magic, RelativeDistinguishedName):
                 attributeTypesAndValues = magic.split()
-            elif isinstance(magic, six.string_types):
-                stringValue = magic
+            elif isinstance(magic, (six.binary_type, six.text_type)):
+                stringValue = _to_bytes(magic)
```
adiroiban: this should be Unicode :)

ldaptor/protocols/pureber.py
```diff
@@ -126,21 +126,29 @@ def __init__(self, tag=None):
...
 
-        return str(self) != str(other)
+        return bytes(self) != bytes(other)
 
     def __hash__(self):
-        return hash(str(self))
+        return hash(bytes(self))
+
+    def __bytes__(self):
+        raise NotImplementedError()
```
adiroiban: why not implemented?

GrayAn:
@adiroiban , thanks a lot for all the remarks! I think I should close this pull request as there are many small changes I need to made, better to show them in a new small pull requests. [This one](https://github.com/twisted/ldaptor/pull/115) is the first.

It looks like using `toWire` method is better than `__str__` and `__bytes__` as Python 2 interpreter doesn't use `__bytes__` method. I don't need to make different logic for different interpreters in these methods.

PR #116, opened 2018-10-12
Switching distinguishedname classes to using bytes

GrayAn:
Greetings!

This is the next small Python 3 compatibility update. ``distinguishedname`` classes now use unicode for storing their arguments and utilize ``toWire`` method to convert them to byte representation. Their ``__str__`` methods were made aliases of ``toWire`` methods with deprecation warnings (as in pureber and pureldap modules).

``ldaptor.encoder.WireStrAlias`` class was created as a base or mixin class for showing deprecation warnings for all these classes.

``DistinguishedName.getDomainName`` method now returns unicode. Can it be a compatibility problem or should I force it to return bytes? The tests for this method are passing without changing because 'dc=example,dc=com' == u'dc=example,dc=com' for Python 2.

Thanks to @adiroiban for the remarks in [the old pull request](https://github.com/twisted/ldaptor/pull/114), I tried to make the appropriate changes in this one.

### Contributor Checklist:

* [X] I have updated the release notes at `docs/source/NEWS.rst` 
* [X] I have updated the automated tests.
* [X] All tests pass on your local dev environment. See `CONTRIBUTING.rst`.


psi29a:
LGTM... what do you think @adiroiban and @cwaldbieser ?

ldaptor/delta.py
```diff
@@ -129,7 +129,7 @@ def __init__(self, dn, modifications=[]):
 
     def asLDIF(self):
         r = []
-        r.append(ldif.attributeAsLDIF('dn', str(self.dn)))
+        r.append(ldif.attributeAsLDIF('dn', self.dn.toWire().decode('utf-8')))
```
adiroiban: This is fine... the internal decoding is a bit strange... but this is a direct translation.

I was expecting for `ldif` to have a method which can taked a name and  DN and add it directly without the need of serialization and deserialization... something like


```suggestion
        r.append(ldif.add('dn', self.dn))
```

ldaptor/delta.py
```diff
@@ -138,7 +138,7 @@ def asLDIF(self):
 
     def asLDAP(self):
         return pureldap.LDAPModifyRequest(
-            object=str(self.dn),
+            object=self.dn,
```
adiroiban: this looks like a change in the API.... it I am misreading this from the diff.

in previous version it was taking bytes and not it takes a DN object?

ldaptor/encoder.py
```diff
@@ -11,3 +13,23 @@ def to_bytes(value):
     if isinstance(value, six.text_type):
         return value.encode('utf-8')
     return bytes(value)
+
+
+def to_unicode(value):
```
adiroiban: Do we need it as a public API? (sorry for missing this comment about `to_bytes`)

I think that we can keep it as a private API (name it with a _)

if it should only be used inside ldaptor it should be private (starting with underscore) and have doctrings :)

No need for dedicated tests (assuming that we have tests for the public API also covering this code)... but all public API should have tests.

Documentation is still nice to have on all code, public or private :)

ldaptor/encoder.py
```diff
@@ -11,3 +13,23 @@ def to_bytes(value):
...
+    if isinstance(value, six.binary_type):
+        return value.decode('utf-8')
+    return value
+
+
+class WireStrAlias(object):
```
adiroiban: missing docstring/documentation for this class.

When, why and how to use this class :)

I saw the comment in the PR description, but people are reading the code :)

In an ideal case, the PR description will be empty and the code and tests will speak for the changes :)

ldaptor/entry.py
```diff
@@ -116,7 +116,7 @@ def __str__(self):
                 vs = list(values)
                 vs.sort()
                 a.append((key, vs))
-        return ldif.asLDIF(self.dn, a)
+        return ldif.asLDIF(self.dn.toWire().decode('utf-8'), a)
```
adiroiban: hm... is this a change in the API?

previously this was distinguishedname.DistinguishedName and now is Unicde... or this is what it looks

ldaptor/protocols/ldap/distinguishedname.py
```diff
@@ -63,59 +51,47 @@ def unescape(s):
...
+    return r
 
 
 class InvalidRelativeDistinguishedName(Exception):
     """Invalid relative distinguished name."""
 
     def __init__(self, rdn):
         Exception.__init__(self)
-        self.rdn = rdn
+        self.rdn = rdn.encode('utf-8') if six.PY2 else rdn
```
adiroiban: Can we have this it was passed in Py2.

My fear is that if someone is already using this on Py2 with UTF-8 it will pass already encoded UTF-8 and then the code will break.

So instead of looking for Python version we can look for input type... if is not 

Also, as part of this work it would be nice to update the docstring to document the input types.

What is the type of `rdn` ? bytes or Unicode?

ldaptor/protocols/ldap/ldapsyntax.py
```diff
@@ -278,7 +278,7 @@ def __hash__(self):
         return hash(str(self))
 
     def bind(self, password):
-        r = pureldap.LDAPBindRequest(dn=str(self.dn), auth=password)
+        r = pureldap.LDAPBindRequest(dn=self.dn, auth=password)
```
adiroiban: is this a change in API? from bytes to DN object?

ldaptor/protocols/ldap/ldapsyntax.py
```diff
@@ -244,7 +244,7 @@ def __str__(self):
         for key, values in lst:
             if key != 'objectClass':
                 a.append((key, values))
-        return ldif.asLDIF(self.dn, a)
+        return ldif.asLDIF(self.dn.toWire().decode('utf-8'), a)
```
adiroiban: Chage in API, from DN Object to Unicode?

tox.ini
```diff
@@ -52,7 +52,7 @@ commands =
 
     test: coverage run --rcfile={toxinidir}/.coveragerc -m twisted.trial {posargs:ldaptor}
 
-    ported: trial ldaptor.test.test_attributeset ldaptor.test.test_autofill ldaptor.test.test_autofill_posix ldaptor.test.test_autofill_samba ldaptor.test.test_config ldaptor.test.test_connector ldaptor.test.test_delta ldaptor.test.test_dns ldaptor.test.test_entry ldaptor.test.test_fetchschema ldaptor.test.test_match ldaptor.test.test_schema ldaptor.test.test_usage ldaptor.test.test_ldapfilter ldaptor.test.test_smbpassword ldaptor.test.test_ldapsyntax ldaptor.test.test_pureber ldaptor.test.test_pureldap
+    ported: trial ldaptor.test.test_attributeset ldaptor.test.test_autofill ldaptor.test.test_autofill_posix ldaptor.test.test_autofill_samba ldaptor.test.test_config ldaptor.test.test_connector ldaptor.test.test_delta ldaptor.test.test_distinguishedname ldaptor.test.test_dns ldaptor.test.test_entry ldaptor.test.test_fetchschema ldaptor.test.test_match ldaptor.test.test_schema ldaptor.test.test_usage ldaptor.test.test_ldapfilter ldaptor.test.test_smbpassword ldaptor.test.test_ldapsyntax ldaptor.test.test_pureber ldaptor.test.test_pureldap
```
adiroiban: this line is hard to check... maybe as a rule, we should only add at the end :)

ldaptor/test/test_encoder.py
```diff
@@ -27,3 +27,10 @@ def test_unicode_object(self):
     def test_bytes_object(self):
         obj = b'bytes'
         self.assertEqual(to_bytes(obj), b'bytes')
+
+
+class WireStrAliasTests(unittest.TestCase):
+
+    def test_not_implemented(self):
```
adiroiban: it would be nice to have a docstring describing this test scenario.
What do you try to test here.

There are no tests for the test code, so human understanding is the best we can get.

The test name and docstring should be explicit about its scope/target/purpose.

something like

```suggestion
    def test_toWire_not_implemented(self):
        """"
        toWire is an abstract method and raises `NotImplementedError`
        """
    
```

ldaptor/encoder.py
```diff
@@ -11,3 +13,23 @@ def to_bytes(value):
...
+        warnings.simplefilter('always', DeprecationWarning)
+        warnings.warn('{0}.__str__ method is deprecated and will not be used '
+                      'for getting bytes representation in the future '
+                      'releases, use {0}.toWire instead'.format(self.__class__.__name__),
+                      category=DeprecationWarning)
+        warnings.simplefilter('default', DeprecationWarning)
+        return self.toWire()
+
+    def toWire(self):
+        raise NotImplementedError()
```
adiroiban: It would be nice to add an helper text to this exception.
That is not instantiate an empty NotImplementedError but instead provide a clue.
For me, it makes it easier to detect what is wrong, without having to search for the traceback.

```suggestion
        raise NotImplementedError('toWire not implemented')
```

ldaptor/test/test_distinguishedname.py
```diff
@@ -315,29 +315,38 @@ def testMalformed(self):
...
-        self.assertEqual(str(d), 'dc=example,dc=com')
+        self.assertEqual(d.toWire(), b'dc=example,dc=com')
 
     def testDN(self):
         proto=dn.DistinguishedName('dc=example,dc=com')
         d=dn.DistinguishedName(proto)
-        self.assertEqual(str(d), 'dc=example,dc=com')
+        self.assertEqual(d.toWire(), b'dc=example,dc=com')
+
+    def testEqualToByteString(self):
```
adiroiban: I know that existing tests don't have docstring, but please try to add a docstring describing the scope/purpose of this test.

in this way, the test can be a documentation for a specific behavoiour of the equal function.

ldaptor/test/test_distinguishedname.py
```diff
@@ -315,29 +315,38 @@ def testMalformed(self):
...
+        d = dn.DistinguishedName('dc=example,dc=com')
+        assert d == b'dc=example,dc=com'
```
adiroiban: I know that previous code was using `assert` but I think that using self.assertEqual() will get a better output in the case of errors and will make future work easier

psi29a:
Going once...

ldaptor/delta.py
GrayAn: True. Appropriate changes were made in ``ldif`` module functions, they now convert passed arguments to their bytes representation using ``to_bytes``.

ldaptor/delta.py
GrayAn: I could use ``self.dn.toWire()`` in this case (and other similar cases) but ``LDAPModifyRequest`` converts ``object`` to its bytes representation when its ``toWire`` method is called (or converted to string when ``__str__`` method was used). So we can pass ``self.dn`` as it is :)

ldaptor/encoder.py
GrayAn: Hmm... Will it be okay if private ``_to_bytes`` and ``_to_unicode`` functions be imported in other ldaptor modules? I have not found such cases in the library yet. Anyway I added the docstrings, thanks for reminding about them :)

ldaptor/encoder.py
GrayAn: Thanks! The description was added.

ldaptor/entry.py
GrayAn: With the changes in ``ldif`` module ``self.dn`` is passed to ``asLDIF`` function as in the previous versions like in ``delta`` module above.

ldaptor/protocols/ldap/ldapsyntax.py
GrayAn: The same idea as in ``delta`` module above: ``LDAPBindRequest`` converts its arguments to byte strings.

ldaptor/protocols/ldap/ldapsyntax.py
GrayAn: Reverted the changes as in the cases above: ``ldif`` module functions convert ``self.dn`` to byte string.

tox.ini
GrayAn: Oh, I will follow it in the future pull requests :)

ldaptor/test/test_encoder.py
GrayAn: Thanks, forgot about it! Docstrings were added

ldaptor/encoder.py
GrayAn: Got it, fixed!

ldaptor/test/test_distinguishedname.py
GrayAn: Docstrings were added for ``testEqualToByteString`` and ``testEqualToString`` tests

ldaptor/test/test_distinguishedname.py
GrayAn: You are right. I was afraid that it will not work this way but didn't check it. Fixed!

GrayAn:
Thanks for the review! I made the most of needed changes but private API question is still actual. Do we need to name modules or functions with API which are considered private with underscores? Do we need to write tests for them? In the latest commits I made changes to ``ldif`` module and it turned out it was not covered by tests — as shown in codecov and tox reports. But it is also looks like "private API" as it is used in limited number of internal modules so maybe it doesn't require tests.

adiroiban:
> Do we need to name modules or functions with API which are considered private with underscores?

If everything from a module should be private, then the module should start with underscore and functions/classes from it don't need to start with an underscore.

If only parts of the module are private, then only those parts should start with the underscore and the module should not start with an underscore.

This is my understanding.

We can start with ldaptor._encoder  and if inside it we will at some point find useful public API we can rename it.

since there was no release yet, we can still rename ldaptor._encoder

------------

> Do we need to write tests for them?

My view is that you don't need to write explicit tests for the private "helpers", but the code should be covered by tests written for the public API.

---------------

> But it is also looks like "private API" as it is used in limited number of internal modules so maybe it doesn't require tests.

Is hard to say how ldaptor is used in the wild.
As soon as a function is public or there is no explic docstring that it should not be used, then we should assume that someone is using it, and if we change the API we will make someone suffer :)

If you think that some API should be private, I think that it is worth creating a separate PR and start by raising a warning.

But, my view is that all code should have at least 100% coverage.

The private API don't need dedicated tests and tests should be written without fiddling with the private code.
In this way, you can refactor the internal implementation and private functions without having the refactor the tests... and in this way the existing tests server as a proof that no regressions were made.

so codecov/patch should be at least 100%, that is all code should be executed at least once

-----------

Now, I know that all this might be considered too much work for a PR and I think that the quality level should be set by the community/users.

I try to follow `Don't do unto others what you don't want done unto you.`

I don't like upgrading a library and having things stop working without any warning :)

-----------

Let me know if you need help with writing the missing tests.

Does it make sense?

ldaptor/encoder.py
adiroiban: Is ok to use private functions inside ldaptor.
The private is a sign to external ldaptor users that they should not use that code...or should use it at their own risk... but we will not warn them when that API is changed.

ldaptor/protocols/ldap/distinguishedname.py
adiroiban: > `rdn` is `str` :)

better to have this answered in the docstring :)

ldaptor/protocols/ldap/ldapsyntax.py
adiroiban: were these changes reverted?

ldaptor/test/test_distinguishedname.py
adiroiban: see https://jml.io/pages/test-docstrings.html for an excellent starting point for writing test docstrings

ldaptor/test/test_encoder.py
```diff
@@ -6,24 +6,50 @@
 
 import six
 
-from ldaptor.encoder import to_bytes
+from ldaptor.encoder import to_bytes, WireStrAlias
 
 
 class WireableObject(object):
+    """
+    Dummy object with bytes representation
```
adiroiban: Instead of using "dummy" maybe is better to describe what it actually does.

Like, `always returns the same value for toWrite` or something like that :)

Or `Implementation with constant toWire value`

ldaptor/encoder.py
GrayAn: So, ``encoder`` module was made private

ldaptor/protocols/ldap/distinguishedname.py
GrayAn: Oh, made more verbose docstring for this exception class :)

ldaptor/protocols/ldap/ldapsyntax.py
GrayAn: My bad, fixed it with the latest commits

ldaptor/test/test_distinguishedname.py
GrayAn: Thanks! Made appropriate changes in docstrings, hope I got the point

ldaptor/test/test_encoder.py
GrayAn: Made more verbose docstring for it

tox.ini
adiroiban: I think that we can also have test_encoder here :)

adiroiban:
> It looks that we have no way of testing them using public API. Should I write a test module for `ldif`? Doesn't look like much work. Maybe I made a mistake when refactored those unused lines of code :)

I think that you did the right thing.

This is a sign that something is not ok with that code.

I think that `header` can be made private but `manyAsLDIFF` looks like a useful code... is you can write tests for that it would be super great... but otherwise I think that we can merge it as it is.

------------

```diff
- r.append(ldif.attributeAsLDIF('dn', str(self.dn)))
+ r.append(ldif.attributeAsLDIF('dn', self.dn))
```

Since `attributeAsLDIF` continues to accept bytes  this is fine.

-----------

It would be great to have tests for ldiff.py (including attributeAsLDIF with bytes and DN as argument)
but I think that we can merge this PR as it is... and add more tests later.

So, if you want to write tests for ldiff, great, if not, let us know and we can look at merging this even if it does not have 100% coverage

many thanks for your help!

tox.ini
GrayAn: Right, added ``test_encoder`` and ``test_ldif``

GrayAn:
Thanks! I implemented ``test_ldif`` module, the coverage looks fine now.

adiroiban:
Thanks @GrayAn . You are a hero!

psi29a:
Thank you both! :)

PR #126, opened 2018-11-08
Python 3 support for ldiftree module

GrayAn:
Greetings!

This is another Python 3 compatibility fix. I had to override `add` and `remove` methods of `LDAPAttributeSet` for it: they now check for existence of a key in a set as a byte or a unicode string. Other updates are mostly usual: explicit using of bytes and switching to `toWire` method from `__str__`.

### Contributor Checklist:


GrayAn:
Codecov complained about some uncovered lines. They were not covered before my changes but I wrote some tests for them anyway :)

GrayAn:
@psi29a Many thanks! I'll do it with the next pull request I think :)

PR #132, opened 2018-11-27
Python 3 compatibility fix for schema and fetchschema modules

GrayAn:
Greetings!

While fixing quick start example to make it Python 3 compatible I found some inconsistencies in `ldapsyntax` module. But before I can fix them I need to make its dependent `fetchschema` and `schema` modules Python 3 compatible as they were still using strings instead of explicit byte strings. Additional test cases for `test_schema` were implemented as the changes in `schema` module affected some lines not covered by tests.

### Contributor Checklist:

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -22,7 +22,8 @@ Changes
 - ``toWire`` method is used to get bytes representation of ``ldaptor.protocols.pureber``,
   ``ldaptor.protocols.pureldap``, ``ldaptor.protocols.ldap.distinguishedname``,
   ``ldaptor.protocols.ldap.ldaperrors``, ``ldaptor.protocols.ldap.ldapclient``,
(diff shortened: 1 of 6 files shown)


psi29a:
Looks like patch is hungup on 96.6%

GrayAn:
My first thought was that all occurrences of `str()` were replaced to `toWire()` for internal classes and the deprecation warning for `__str__` method was not used anywhere. I wrote an additional test for `test_encoder` that checks `DeprecationWarning` is emitted. But codecov is still highlight the lines of this method.

Now I noticed that [codecov is actually warns](https://docs.codecov.io/docs/viewing-source-code#section-line-coverage-changes) that coverage for this method was increased. Maybe I understand it wrong...

GrayAn:
It looks like I found what bothered codecov. I close this PR and make a new one.

Issue #144, opened 2019-04-09
value.toWire raises NotImplementedError


donalm:
The 'quick usage example' on ldaptor's github homepage works for me (with minimal obvious modifications) on version 16 of ldaptor on Python 2.7

For version 19 of ldaptor, I get this error with python 3.6.5
```python
Unhandled Error
Traceback (most recent call last):
  File "/mnt/software/external/python_extras/0.8.0/platform-linux/python-3.6/lib/python3.6/site-packages/twisted/python/log.py", line 103, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/mnt/software/external/python_extras/0.8.0/platform-linux/python-3.6/lib/python3.6/site-packages/twisted/python/log.py", line 86, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/mnt/software/external/python_extras/0.8.0/platform-linux/python-3.6/lib/python3.6/site-packages/twisted/python/context.py", line 122, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/mnt/software/external/python_extras/0.8.0/platform-linux/python-3.6/lib/python3.6/site-packages/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/mnt/software/external/python_extras/0.8.0/platform-linux/python-3.6/lib/python3.6/site-packages/twisted/internet/posixbase.py", line 614, in _doReadOrWrite
    why = selectable.doRead()
  File "/mnt/software/external/python_extras/0.8.0/platform-linux/python-3.6/lib/python3.6/site-packages/twisted/internet/tcp.py", line 243, in doRead
    return self._dataReceived(data)
  File "/mnt/software/external/python_extras/0.8.0/platform-linux/python-3.6/lib/python3.6/site-packages/twisted/internet/tcp.py", line 249, in _dataReceived
    rval = self.protocol.dataReceived(data)
  File "/mnt/software/external/python_extras/0.8.0/platform-linux/python-3.6/lib/python3.6/site-packages/ldaptor/protocols/ldap/ldapclient.py", line 70, in dataReceived
    if not o:
  File "/mnt/software/external/python_extras/0.8.0/platform-linux/python-3.6/lib/python3.6/site-packages/ldaptor/protocols/pureber.py", line 132, in __len__
    return len(self.toWire())
[...]

PR #145, opened 2019-04-10
LDAPSearchResultReference basic implementation

GrayAn:
This PR is a fix for issue #144 . `LDAPSearchResultReference` class was not fully implemented and did not have tests before (so I accidently broke it while porting `pureldap` code for Python 3). I attempted to implement it accordingly to the [RFC](https://tools.ietf.org/html/rfc4511#section-4.5.3) but not sure if it really works: I need a LDAP server with this feature to test it (Active Directory for example).

### Contributor Checklist:

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -4,11 +4,17 @@ Changelog
 Release 19.1 (Unreleased)
 -------------------------
 
+Features
+^^^^^^^^
+
+- Basic implementation of ``ldaptor.protocols.pureldap.LDAPSearchResultReference``.
+
 Bugfixes
 ^^^^^^^^
 
 - ``DeprecationWarning`` stacklevel was set to mark the caller of the deprecated
   methods of the ``ldaptor._encoder`` classes.
+- ``NotImplementedError`` for ``ldaptor.protocols.pureldap.LDAPSearchResultReference`` was fixed.
 
 Release 19.0 (2019-03-05)
 -------------------------
--- a/ldaptor/protocols/pureldap.py
+++ b/ldaptor/protocols/pureldap.py
@@ -226,20 +226,38 @@ class LDAPReferral(BERSequence):
     tag = CLASS_CONTEXT | 0x03
 
 
-# This is currently just a stub and implements no real functionality.
-class LDAPSearchResultReference(LDAPProtocolResponse):
+class LDAPBERDecoderContext_LDAPSearchResultReference(BERDecoderContext):
+    Identities = {
+        BEROctetString.tag: LDAPString,
+    }
+
+
+class LDAPSearchResultReference(LDAPProtocolResponse, BERSequence):
     tag = CLASS_APPLICATION | 0x13
 
-    def __init__(self):
+    def __init__(self, uris=None, tag=None):
         LDAPProtocolResponse.__init__(self)
+        BERSequence.__init__(self, value=[], tag=tag)
+        assert uris is not None
+        self.uris = uris
 
     @classmethod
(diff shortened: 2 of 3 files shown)


GrayAn:
Well, this time I have no idea what codecov does not like :)

psi29a:
Can anyone who has access to an AD, can they test this code against it?

cwaldbieser:
I'll try to take a look at it.

cwaldbieser:
Hmm-- I'm not sure I can generate a query that will create an LDAP Search Result reference.  I do have access to an AD, but it is just  a single, unpartitioned domain.  Is there some way to create an reference entry that can be queried?

On a different note, I ran into a number of cases where I got an error similar to the one in #144.

> During handling of the above exception, another exception occurred:

It looks like all the exceptions being raised inherit from `ldaperrors.LDAPException`.  That class has had its `__str__()` implementation switched to `toWire()`, but that doesn't seem right-- the exception handling code in Twisted seems to expect an exception that can be printed:

      File "/home/waldbiec/.virtualenvs/txldap_ad_client-hpiTWo4R/lib/python3.6/site-packages/twisted/python/reflect.py", line 448, in safe_str
        return str(o)
    TypeError: __str__ returned non-string (type bytes)

I got this error during StartTLS negotiation and when I tried to query the rootDSE and got a "NO SUCH OBJECT" response.  In the former case, the issue seems to be similar to the 2nd error described in #144.

In the `protocols.ldap.ldapclient.LDAPClient._cbStartTLS()` callback, the comparison between `msg.responseName` and `pureldap.LDAPStartTLSResponse.oid` fails because one is a byte-string and the other is a unicode string.  If I decoded `msg.responseName` to UTF-8 first, the comparison worked as I would have expected.

I only tried this in python 3.  I've been weaning myself off python 2.7 this year.

GrayAn:
Thanks @cwaldbieser !

I fixed these issues in the other pull requests: #146 and #147 .

GrayAn:
One of my colleagues gave me an access to his Active Directory test instance. And luckily it responded with a `LDAPSearchResultReference` message to the search request:
```
C->S LDAPMessage(id=2, value=LDAPSearchRequest(baseObject=u'cn=configuration,dc=foo,dc=kama,dc=gs', scope=2, derefAliases=0, sizeLimit=0, timeLimit=0, typesOnly=0, filter=LDAPFilter_present(value='objectClass'), attributes=('objectClass',)), controls=None)
C<-S LDAPMessage(id=2, value=LDAPSearchResultEntry(objectName='CN=Configuration,DC=foo,DC=kama,DC=gs', attributes=[('objectClass', ['top', 'configuration'])], controls=None)
...
C<-S LDAPMessage(id=2, value=LDAPSearchResultReference(uris=[LDAPString(value='ldap://foo.kama.gs/CN=Schema,CN=Configuration,DC=foo,DC=kama,DC=gs')]), controls=None)
C<-S LDAPMessage(id=2, value=LDAPSearchResultDone(resultCode=4), controls=None)
```
I tried it on Python 2.7 and 3.6 versions.
Looks like it actually works, not only in the tests :)

cwaldbieser:
Thanks, this looked good to me.  It's good to know this actually worked against a real example!

GrayAn:
Thank you! Maybe we can make a release? There was a number of [bug fixes](https://github.com/twisted/ldaptor/blob/master/docs/source/NEWS.rst#release-191-unreleased) since the last release.

psi29a:
Glad this was tested against something live.

Is everyone on board for another release?

cwaldbieser:
Yes, a release seems prudent.

PR #159, opened 2019-12-19
Replace depricated calls to decodestring with decodebytes

mbish:
This change is nominal and removes a warning message about the use of deprecated methods. 

### Contributor Checklist:

* [?] I have updated the automated tests.

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -9,6 +9,7 @@ Features
 
 Changes
 ^^^^^^^
+- Replaced depricated calls to base64.decodestring with base64.decodebytes
 
 Bugfixes
 ^^^^^^^^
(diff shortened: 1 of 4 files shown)


GrayAn:
Hi @mbish ! Unfortunately it breaks backward compatibility with Python 2.7 as the `decodebytes` function was introduced in Python 3.1 and was not backported. Ldaptor has received its Python 3 support only this year so it would be great if it will work with 2.7 for a while though 2.7 support will end soon.

I suggest you to create a compatibility function like this:
```
try:
    decodebytes = base64.decodebytes
except AttributeError:
    decodebytes = base64.decodestring
```
or
```
import six

decodebytes = base64.decodebytes if six.PY3 else base64.decodestring
```
It can be stored in `ldaptor._encoder` module with the other decoding helper functions.

PR #161, opened 2020-02-18
avoid duplicating ldaptor code in LDAPServerWithUPNBind

graingert:
### Contributor Checklist:

* [ ] I have updated the automated tests.

### DIFF
--- a/docs/source/examples/ldaptor_with_upn_bind.py
+++ b/docs/source/examples/ldaptor_with_upn_bind.py
@@ -18,80 +18,45 @@ userPrincipalName: bob@ad.example.org
 """
 from __future__ import absolute_import
 
-from ldaptor import interfaces
 from ldaptor.protocols import pureldap
-from ldaptor.protocols.ldap import distinguishedname, ldaperrors
 from twisted.internet import defer
 from ldaptor.protocols.ldap.ldapserver import LDAPServer
 
 
-class LDAPServerWithUPNBind(LDAPServer):
+class LDAPServerWithUPNBind(LDAPServer, object):
     """
     An LDAP server which support BIND using UPN similar to AD.
     """
-    _loginAttribute = b'userPrincipalName'
 
-    def handle_LDAPBindRequest(self, request, controls, reply):
-        if request.version != 3:
-            raise ldaperrors.LDAPProtocolError(
-                'Version %u not supported' % request.version)
+    _loginAttribute = b"userPrincipalName"
 
-        self.checkControls(controls)
-
-        if request.dn == b'':
-            # anonymous bind
-            self.boundUser = None
-            return pureldap.LDAPBindResponse(resultCode=0)
-
-        root = interfaces.IConnectedLDAPEntry(self.factory)
-
-        def _gotUPNResult(results):
-            if len(results) != 1:
-                # Not exactly one result, so this might not be an UNP.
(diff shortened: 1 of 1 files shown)


adiroiban:
Thanks for the cleanup.
@graingert  let us know when you think this is ready for review :)

graingert:
@adiroiban I think it's ready to go

docs/source/examples/ldaptor_with_upn_bind.py
```diff
@@ -20,78 +20,44 @@
...
-            # A single result, so the UPN might exist.
-            return results[0].dn
-
-        if b'@' in request.dn and b',' not in request.dn:
+    @defer.inlineCallbacks
+    def handle_LDAPBindRequest(self, request, *args, **kwargs):
+        @defer.inlineCallbacks
+        def _request():
+            if not (b"@" in request.dn and b"," not in request.dn):
+                defer.returnValue(request)
```
adiroiban: Maybe something like this

```suggestion
        def _resolveUPNBindDN():
           """
           See if we have an MS UPN style BIND DN request and resolve it to the
           actual LDAP BindDN.
           """
            if b"@" not in request.dn or b"," in request.dn):
                # This is not an UPN bind request
                defer.returnValue(request)
```

docs/source/examples/ldaptor_with_upn_bind.py
```diff
@@ -20,78 +20,44 @@
...
+            root = interfaces.IConnectedLDAPEntry(self.factory)
             # This might be an UPN request.
```
adiroiban:
```suggestion
            # The the LDAP DN associated with the UPN request.
```

docs/source/examples/ldaptor_with_upn_bind.py
```diff
@@ -20,78 +20,44 @@
...
-                return msg
-            d.addCallback(_cb)
-            return d
-        d.addCallback(_gotEntry, request.auth)
-
-        return d
+            filter_text = b"(" + self._loginAttribute + b"=" + request.dn + b")"
+            results = yield root.search(filterText=filter_text)
+            if len(results) != 1:
+                defer.returnValue(request)
```
adiroiban:
```suggestion
            if len(results) != 1:
                # We could not find an exact UPN match so we go with the
                # requested Bind DN.
                defer.returnValue(request)
```

adiroiban (changes requested):
Thanks for the cleanup.

Please see my comments.

Even thou I wrote this code a few months ago, I feel that it needs more documentation.
With the current version, at the first read it was hard to understand what is going on there and why we have all that logic.

psi29a:
@graingert will you pick this work back up?

graingert:
Sorry, I no longer work for the company that was using ldaptor so I probably won't pick this up

adiroiban:
@psi29a do you have time to review this code ?
I am happy to work on implementing any changes.

---------

I think that the code from here is quite useful, when you want to use ldaptor to pretent it is an AD LDAP server.

Thanks

psi29a:
> @psi29a do you have time to review this code ? I am happy to work on implementing any changes.
> 
> I think that the code from here is quite useful, when you want to use ldaptor to pretent it is an AD LDAP server.
> 
> Thanks

I don't have an AD LDAP server to verify against, but perhaps I can clean this up a bit. Would you be able to test and validate?

PR #164, opened 2020-02-19
Add a repr for LDAPBindResponse_serverSaslCreds so that binary data can be logged

xdesai:
In practice the `LDAPBindResponse_serverSaslCreds` object has a `value` that is not a utf-8 string like most objects, but is instead binary data.
This means that when you try to `repr()` the message for logging you can't use the default `repr` that is on `BEROctetString` because it tries to do a utf-8 decode.

Since the `LDAPBindResponse_serverSaslCreds` object already existed I just implemented a `__repr__` on it that didn't try to do utf-8 decoding.

As far as testing goes I was able to write a couple unit tests for this functionality and we also are successfully using this patch in our project that uses ldaptor. Due to the default nature of strings in python 2 vs python 3 I could not write a test case that passed in both language versions without using six. If someone knows of a better way please feel free to let me know or just edit the PR directly.

Let me know if you have any other questions or feedback!
Xander

### Contributor Checklist:

* [X] I have updated the automated tests.
* [X] All tests pass on your local dev environment. See `CONTRIBUTING.rst`.

### DIFF
--- a/ldaptor/test/test_pureldap.py
+++ b/ldaptor/test/test_pureldap.py
@@ -1571,8 +1571,8 @@ class TestRepresentations(unittest.TestCase):
         self.assertEqual(actual_repr, expected_repr)
 
     def test_ldap_bind_response_server_sasl_creds_with_tag_repr(self):
-        """ ServerSaslCreds will often have binary data. A custom repr is needed because
-        it cannot be turned into a unicode string like most BEROctetString objects.
+        """ An LDAPBindResponse_serverSaslCreds with a non-standard tag will have that
+        tag show up in the text representation.
         """
         sasl_creds = pureldap.LDAPBindResponse_serverSaslCreds(value=b'NTLMSSP\xbe', tag=12)
         if six.PY3:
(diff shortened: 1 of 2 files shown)


ldaptor/test/test_pureldap.py
```diff
@@ -1561,3 +1561,29 @@ def test_matching_rule_assertion_with_tag_repr(self):
...
+            expected_repr = r"LDAPBindResponse_serverSaslCreds(value=b'NTLMSSP\xbe')"
+        else:
+            expected_repr = "LDAPBindResponse_serverSaslCreds(value=NTLMSSP\xbe)"
+
+        actual_repr = repr(sasl_creds)
+        self.assertEqual(actual_repr, expected_repr)
+
+    def test_ldap_bind_response_server_sasl_creds_with_tag_repr(self):
+        """ ServerSaslCreds will often have binary data. A custom repr is needed because
+        it cannot be turned into a unicode string like most BEROctetString objects.
```
adiroiban: this docstring looks like a duplicate.
I don't think that using duplicate docstring add any value... it can reduce the value as it can cause confusion.


Maybe use something like this: A response with a non-standard tag will have the tag visible in the text representation.

adiroiban:
I have updated to branch..and I hope this will fix the test coverage.

btw.. excellent PR description :) thanks for that!

ldaptor/test/test_pureldap.py
xdesai: Good feedback! I'll update the diff :)

adiroiban:
I think that this is an internal thing and we don't need a release notes.

I would say that `__repr__` should be considered private API ... that is, people can use it, but we should not make any backward compatibility commitment  and people should expect things to break if repr output is used for machine processing

I will merge is at it is and will look at the missing test coverage in a dedicated PR.

Thanks again!

PR #175, opened 2020-09-22
Release to pypi on a new tag via github-action code.

adiroiban:
# Scope

This add automatic release when a new tag is pushed.

The release is done via GitHub actions 

# Changes

Add a Github Action workflow based on https://github.com/pypa/gh-action-pypi-publish

Use the PyPi token generated by @psi29a 

# How to test

All automated tests are executed before a release.

There is no actual PyPI release if this is a not a tag. 

Every test matrix run uploads their dist as an artifact

Testing might be a trial and error and we might need to manually delete some packages.


pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[build-system]
```
graingert, line 1: To switch to flit you'll need to enable tox isolated_build and delete the setup.py:

https://tox.readthedocs.io/en/latest/config.html

graingert: You may prefer to use:

```
pipx run setup-py-upgrade .
```

and then use the setuptools build-system

tox.ini
```diff
@@ -67,3 +67,14 @@ commands =
     mkdir -p docs/build
     sphinx-build -b html -Ean -j 2 -w build/sphinx-errors \
         docs/source docs/build
+
+[testenv:release]
+deps =
+    toml
+    pep517
+    flit
```
graingert, line 75: You don't need flit and pep517 - pep517.build will download the build-system on the fly

adiroiban:
@graingert I have added you to this repo. You should now be able to push changes to this branch. I am new to the while python3 new stuff :)

@psi29a I hope you are ok with this. You can review the current repo access at https://github.com/twisted/ldaptor/settings/access


At the same time, I don't know how we can test this.. .I guess that we will need to push some testing versions and see that all is ok and then push the final version.
We can later delete the pre-release version.

graingert:
closing in favour of https://github.com/twisted/ldaptor/pull/178

PR #176, opened 2020-09-22
configure pep517, and configure tox to build and test pep517 built wheels

graingert:
I see in https://github.com/twisted/ldaptor/pull/175 a desire to test the released artefact as part of the release. This configures tox to create a .whl using pep517 file and then test that.

### Contributor Checklist:


adiroiban:
Thanks for the update.

Can you please also add the `tox -e release` environment and update CONTRIBUTING.rst to explain how to create the release files?

In this way, the other PR can be focused only on introducing the GitHub Action part

setup.py
```diff
@@ -25,51 +25,4 @@ def find_version(*file_paths):
...
-                   "bin/ldaptor-ldap2passwd",
-                   "bin/ldaptor-getfreenumber",
-                   "bin/ldaptor-ldap2pdns",
-                   "bin/ldaptor-find-server",
-                   "bin/ldaptor-rename",
-                   "bin/ldaptor-fetchschema",
-                   "bin/ldaptor-ldifdiff",
-                   "bin/ldaptor-ldifpatch"]
-          )
+    setup(version=find_version("ldaptor", "__init__.py"))
```
adiroiban, line 28: do we need this here?

My understanding is that when pep571 is used the version is automatically extracted.

graingert: it's still needed - pep517 is all about calling into build systems. the setuptools build system just adds a `setup.py` if absent and calls it

CONTRIBUTING.rst
```diff
@@ -44,6 +44,14 @@ docs/source/NEWS.rst in a wording targeted to end users.
 Try not to write the release notes as a commit message.
 
 
+Release process
+---------------
+
+The release is done manually.
+You can test the release process (without the publish) using `tox -e release`.
+Inspect the released files with `tree dist`, you could upload them with `twine`.
```
graingert, line 52:
```suggestion
Inspect the distributable files with `tree dist`, you could upload them with `twine`.
```

pyproject.toml
```diff
@@ -67,3 +79,15 @@ commands =
...
+
+[testenv:release]
+basepython=python3.8
+deps = pep517
+whitelist_externals =
+    cp
+    rm
+commands =
+    rm -rf {toxinidir}/dist
+    cp --no-target-directory --recursive {distdir} {toxinidir}/dist; # copy the wheel built by tox-wheel
```
graingert, line 91: @adiroiban this is where the magic happens - I can now run `pipx run tox -e py38-test-twlatest,release && twine upload dist/*` and tox will build the whl and *that whl file* will be installed and tested and *that whl file* will be uploaded to pypi

graingert:
> Can you please also add the tox -e release environment and update CONTRIBUTING.rst to explain how to create the release files?

@adiroiban done!

pyproject.toml
adiroiban: I am -1 for any magic stuff :) When I read that line it makes no sense to me.... so I guess that a comment is required.

What is wrong with using the `python -m pep517.build --source --binary --out-dir dist/ .` command to generate the source dist and binary wheel?

Also, I think that `tox -e release` is the wrong environment name. Maybe it should be `tox -e create-dist` or something like that.

The actual release/publish should be a separate step as it required a repo URL + credentials.

graingert: > What is wrong with using the python -m pep517.build --source --binary --out-dir dist/ . command to generate the source dist and binary wheel?

I want to test the actual whl file that gets released

graingert: > Also, I think that tox -e release is the wrong environment name. Maybe it should be tox -e create-dist or something like that.

I'm happy with either name, can you push a fix with what you think is best?

graingert: > @adiroiban this is where the magic happens 

Maybe magic was the wrong word. This line is the culmination/apex of the PR that everything else was building to. It allows us to create an artifact, test it and then distribute it

graingert: @adiroiban pinging you on this change. What sort of comment would be best here?

adiroiban: Hi... I got it now :)
There was the comment that the wheel is created by tox, but I have not seen any tox command to create it.

But I now know what it going on here. For every env, tox will automatically create a wheel and install it in the dedicated virtualenv.

adiroiban (approved):
Look good. Thanks.

You should be able to merge the PR now.

PR #177, opened 2020-09-23
Configure GitHub actions

graingert:
## Remove this paragraph

Please have a look at our dev docs before submitting your PR:
https://github.com/twisted/ldaptor/blob/master/CONTRIBUTING.rst

### Contributor Checklist:

I copied this from modernize

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,6 +15,11 @@ jobs:
       matrix:
         python-version: [3.5, 3.6, 3.7, pypy3, 3.8]
         os: [macOS-latest, ubuntu-latest, windows-latest]
+        exclude:
+          - os: windows-latest
+            python-version: pypy3
+          - os: macOS-latest
+            python-version: pypy3
 
     steps:
       - name: Checkout
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,4 +45,4 @@ jobs:
       - name: tox
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: tox -e py-test-twlatest,py-linters,publishcov,release
+        run: tox -e py-test-twlatest,py-test-twtrunk,py-linters,publishcov,release
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,48 @@
+name: CI
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - v*
+  pull_request:
+
+jobs:
+  tox:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
+        os: [macOS-latest, ubuntu-latest, windows-latest]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set Up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Get pip cache dir
+        id: pip-cache
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+
+      - name: pip cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.cfg') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
+      - name: Install
+        run: |
+          pip install tox
+
+      - name: tox
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: tox -e py-test-twlatest,py-linters,publishcov,release
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,7 +34,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
-          key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.cfg') }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml', 'setup.py', 'setup.cfg') }}
           restore-keys: |
             ${{ runner.os }}-pip-
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8]
+        python-version: [3.5, 3.6, 3.7, pypy3, 3.8]
         os: [macOS-latest, ubuntu-latest, windows-latest]
 
     steps:
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,13 +54,13 @@ before_install:
       source <(curl -SL https://git.io/fhPYW -o -)
     fi
 install:
-  - pip install tox codecov
+  - pip install tox
 
 script:
   - tox -e $TOX_ENV
 
 after_script:
-  - codecov
+  - tox -e publishcov
 
 notifications:
     email: false
--- a/ldaptor/test/test_ldiftree.py
+++ b/ldaptor/test/test_ldiftree.py
@@ -3,9 +3,11 @@ Test cases for LDIF directory tree writing/reading.
 """
 
 import os
+import sys
 import random
 import errno
 import shutil
+import unittest as stdlib_unittest
 
 import six
 from twisted.trial import unittest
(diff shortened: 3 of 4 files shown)


pyproject.toml
```diff
@@ -90,4 +89,11 @@ commands =
     rm -rf {toxinidir}/dist
     cp --no-target-directory --recursive {distdir} {toxinidir}/dist; # copy the wheel built by tox-wheel
     {envpython} -m pep517.build --source --out-dir={toxinidir}/dist {toxinidir}
+
+[testenv:coveralls]
+passenv = GITHUB_*
+deps =
+  coveralls
+  coverage>=5.3
+commands = coveralls
```
graingert, line 98: @adiroiban I used coveralls because that's what I'm familliar with and use in https://github.com/PyCQA/modernize

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Get pip cache dir
+        id: pip-cache
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+
+      - name: pip cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.cfg') }}
```
adiroiban, line 37: I think that ` pyproject.toml` should also be included as a key to the cache... not sure what setup.cfg is :)

I guess that  pyproject.toml is replacing the setup.py and setup.cfg file

pyproject.toml
adiroiban: I am ok with using coveralls .... a long time ago, I was using coveralls , but it was failing so I switched to codecov.

I think that it would be interesting to run both coveralls and codecov in parallel ... but to be done in a separate PR.

Maybe coveralls should be left for a separate PR and for this PR just focus on adding the basic GitHub actions matrix.

adiroiban:
I don't think that needs NEWS update as this is not a public facing change.

.github/workflows/main.yml
graingert: https://github.com/twisted/ldaptor/blob/6432ca1d9078a5439565113e6344212812b39c0f/setup.cfg#L22-L28

adiroiban:
Do we need 3.5, 3.6, 3.7, and 3.8? I guess that just having 3.5 and 3.8 should give us a good coverage.

Or maybe run 3.5 and 3.7 only on Linux... and also run only on linux tests with latest twisted release and with twisted trunk

graingert:
> Do we need 3.5, 3.6, 3.7, and 3.8? I guess that just having 3.5 and 3.8 should give us a good coverage.
> Or maybe run 3.5 and 3.7 only on Linux... and also run only on linux tests with latest twisted release and with twisted trunk

GHA's parallel build limits per project are currently ludicrously high. We can remove some if they ever lower these limits

adiroiban:
> GHA's parallel build limits per project are currently ludicrously high. We can remove some if they ever lower these limits

But I guess that they are high just to kill the competition.
Once GitHub will have a monopoly I expect they will reduce them.

But at the same time, if they don't provide any value, those extra tests are just creating noise for ldaptor and wasting resource for the whole humanity :)

graingert:
>  wasting resource for the whole humanity :)

burning github compute means they will be force to charge ICE more

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+            ${{ runner.os }}-pip-
+
+      - name: Install
+        run: |
+          pip install tox
+
+      - name: tox
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: tox -e py-test-twlatest,py-test-twtrunk,py-linters,publishcov,release
```
adiroiban, line 48: think that py-linters can be executed as part of a separate workflow and it should be enough to run it only on Linux.
Same for the release environment.

My problem with tox in general is that it is very verbose / noisy.

For example, if a tests fails, with the current setup I will have to scroll 2500 lines and somewhere in the middle I will see the error details.

What I do for my projects, is send all important output to a separate file and then print that file as a separate step.
That separate file will contain only the important test failure details.

For tox, as a workaround I am running tox --notest first with full verbosity to create the environmet and have a trace of the exact versions used and then I run tox -qqq to just run the tests and see the errors in a dedicated step.

adiroiban (commented):
just sending my current comments as I am constantly loosing them while the branch is updated while doing a review

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: pip cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml', 'setup.py', 'setup.cfg') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
+      - name: Install
```
adiroiban, line 41:
```suggestion
      - name: Install build dependencies
```

.travis.yml
```diff
@@ -54,13 +54,13 @@ before_install:
...
 install:
-  - pip install tox codecov
+  - pip install tox
 
 script:
   - tox -e $TOX_ENV
 
 after_script:
-  - codecov
+  - tox -e publishcov
```
adiroiban, line 63: do we need this ?

For me spinning a new virtualenv just to push the coverage is wasted time, especially since this code is designed to be executed in Travis-CI and GHA where we have a new fresh VM for each build.

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          pip install tox
+
+      - name: tox
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: tox -e py-test-twlatest,py-linters,publishcov,release
```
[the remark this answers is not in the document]
graingert: I think we can cope with this for now and use https://github.com/ymyzk/tox-gh-actions/issues/32 when it's available

.travis.yml
graingert: this makes everything consistent, and we can start using `tox --parallel-live` soon

.github/workflows/main.yml
graingert:
```suggestion
      - name: Install tox
```

.travis.yml
adiroiban: We can leave this change but a developer running tests on the local system should not push the code coverage results.

graingert: The GITHUB_TOKEN is rolled every build so would be very tricky for someone to succeed in uploading this once the CODECOV_* Travis vars are revoked

adiroiban:
I think that this is a bug in the codecov reporting tool, or we might need to export more vaiables

With GitHub actions is hard to tell which coverage report was pushed by which build

https://codecov.io/gh/twisted/ldaptor/commit/934096d82b3da2a4f4bfc149fa8d702cea45cb53/build

pyproject.toml
```diff
@@ -81,13 +81,20 @@ commands =
...
     cp
     rm
 commands =
     rm -rf {toxinidir}/dist
-    cp --no-target-directory --recursive {distdir} {toxinidir}/dist; # copy the wheel built by tox-wheel
+    cp -r {distdir} {toxinidir}/dist; # copy the wheel built by tox-wheel
     {envpython} -m pep517.build --source --out-dir={toxinidir}/dist {toxinidir}
+
+[testenv:publishcov]
+passenv = CI GITHUB_* TRAVIS TRAVIS_* CODECOV_*
```
adiroiban, line 94: See https://github.com/codecov/codecov-python#using-tox

Maye use `passenv = TOXENV CI TRAVIS TRAVIS_* CODECOV_*`

But I am +1 on just using codecov outside of tox to puslish the coverage.

For me, for coverage publishing, tox just adds an extra layer which is not needed.

ldaptor/test/test_ldiftree.py
```diff
@@ -21,6 +23,18 @@ def writeFile(path, content):
...
 
+skipIfWindowsOrRoot = stdlib_unittest.skipIf(
+    sys.platform == "win32" or os.getuid() == 0,
+    "Can't test on windows or as root",
+)
+
+
+skipIfWindows = stdlib_unittest.skipIf(
+    sys.platform == "win32",
+    "TODO: fails on windows",
```
adiroiban, line 34: I think that this needs a dedicated issue and the TODO: can include the link to that issue.

adiroiban:
The codecov/project/tests is faiing as the report for python 2.7 test run from Travis-CI failed.
You might look at this result and see it as green https://travis-ci.org/github/twisted/ldaptor/jobs/729596973 

but in fact there is this error

```
Error: HTTPSConnectionPool(host='codecov.io', port=443): Max retries exceeded with url: /upload/v4?service=travis&pr=177&yaml=codecov.yml&slug=twisted%2Fldaptor&job=729596973&commit=934096d82b3da2a4f4bfc149fa8d702cea45cb53&branch=master&package=py2.1.9&build=483.1 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7fc8aafea2d0>: Failed to establish a new connection: [Errno 110] Connection timed out',))
```


The `codecov/project/tests` was introduced to make sure that we don't have "dead" code inside the tests. That is we don't have tests that are never executed.

adiroiban (approved):
Once this is green, feel free to merge :)

Great work. Thanks.

And maybe Travis-CI can be removed in this same PR

pyproject.toml
graingert: codecov doesn't collect `GITHUB_` metadata :(

.github/workflows/main.yml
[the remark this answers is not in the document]
graingert: I like running the linters on each build. there's some new syntax in 3.9 and 3.10 and running the lint rule on only those would result in a test failure rather than lint failure

graingert:
> And maybe Travis-CI can be removed in this same PR

I can't remove it because it's a require build

PR #178, opened 2020-09-23
Release to PyPI on a pushed tag using Github Actions

graingert:
# Scope

This add automatic release when a new tag is pushed.

The release is done via GitHub actions 

# Changes

Add a Github Action workflow based on https://github.com/pypa/gh-action-pypi-publish

Use the PyPi token generated by @psi29a 

# How to test

All automated tests are executed before a release.

There is no actual PyPI release if this is a not a tag. 

Every test matrix run uploads their dist as an artifact

Testing might be a trial and error and we might need to manually delete some packages.

### DIFF
--- a/ldaptor/__init__.py
+++ b/ldaptor/__init__.py
@@ -1,3 +1,4 @@
+"""A Pure-Python Twisted library for LDAP"""
 __version__ = "19.1.0"
 
 __title__ = "ldaptor"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,3 +51,28 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: tox -e py-test-twlatest,py-test-twtrunk,py-linters,publishcov,release
+
+      - name: upload dist
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.os }}_${{ matrix.python-version}}_dist
+          path: dist
+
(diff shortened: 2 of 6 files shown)


ldaptor/__init__.py
```diff
@@ -1,3 +1,4 @@
+"""A Pure-Python Twisted library for LDAP"""
 __version__ = "19.1.0"
```
graingert, line 2: I didn't do the version change now - I can make the version change as a new PR

graingert:
@adiroiban can you take a look at this?

ldaptor/__init__.py
adiroiban: I think that as part of this PR we should update to 20.0.0 and do a release just to make sure that it works.

This can be done as 20.0.0.rc1 release which will be triggered by a push (disable tag filtering)

adiroiban:
I don't think that creating the same wheel 13 times and then uploading it is a good idea.

I would prefer that `all-successful` to be renamed as `publish-on-new-tag` and build the source and wheel and push it, and not pull it as a wheel.

graingert:
> build the source and wheel and push it, and not pull it as a wheel.

I disagree, the purpose of this approach is to carry a tested artifact all the way to release

I could limit the release upload one artifact, but I'd just have to undo if when testing reproducible builds

adiroiban (approved):
I don't want to block this.

So, as long as it works, please merge it and trigger a 20.0.0 release.

Thanks!

PR #187, opened 2020-09-29
add new shields.io badges for travis, github, rtd and pypi

graingert:
### Contributor Checklist:

* [ ] I have updated the automated tests.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,10 +1,21 @@
 Ldaptor
 =======
 
-.. image:: https://travis-ci.org/twisted/ldaptor.svg?branch=master
+.. image:: https://img.shields.io/travis/com/twisted/ldaptor?label=TravisCI&logo=travis
+    :alt: Travis (.com)
     :target: https://travis-ci.org/twisted/ldaptor
-.. image:: https://codecov.io/gh/twisted/ldaptor/branch/master/graph/badge.svg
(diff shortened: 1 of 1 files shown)


adiroiban (approved):
I have tested this by checking the readme rendering from the branch at https://github.com/graingert/ldaptor/tree/add-shields

I think that Travis can be removed, as the travis-ci will be removed soon.

Not sure why readthedocs badge is red.

Thanks

psi29a (changes requested):
I agree, if we are going to use github only hooks and no travis, then the travis badge should go.

graingert:
Currently ldaptor requires Travis to test using python v2.7, once that requirement is gone I'll remove all record of Travis in one go

graingert:
> Not sure why readthedocs badge is red.

I posted this elsewhere but for other's reference it was because the epub builds failed which are not run in PRs https://github.com/readthedocs/readthedocs.org/issues/7116

adiroiban:
I think that this can be merges as it is. My comment about travis shield was just a suggestion. 

Travis is still required at this time, as py2 is not yet removed in master.

---------

regarding read the docs... can we disable the epub file generation from read the docs config file? Maybe as a drive-by to this PR?

Thanks!

PR #195, opened 2020-09-30
Drop support for py35

graingert:
### Contributor Checklist:

### DIFF
--- a/ldaptor/__init__.py
+++ b/ldaptor/__init__.py
@@ -7,4 +7,4 @@ __uri__ = "https://github.com/twisted/ldaptor"
 
 __license__ = "MIT"
 __author__ = "The ldaptor developers"
-__copyright__ = f"Copyright (c) 2002-2019 {__author__}"
+__copyright__ = f"Copyright (c) 2002-2021 {__author__}"
--- a/ldaptor/__init__.py
+++ b/ldaptor/__init__.py
@@ -7,4 +7,4 @@ __uri__ = "https://github.com/twisted/ldaptor"
 
 __license__ = "MIT"
 __author__ = "The ldaptor developers"
-__copyright__ = "Copyright (c) 2002-2019 {}".format(__author__)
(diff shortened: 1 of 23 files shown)


graingert:
waiting for https://github.com/twisted/twisted/pull/1386

ldaptor/__init__.py
```diff
@@ -7,4 +7,4 @@
 
 __license__ = "MIT"
 __author__ = "The ldaptor developers"
-__copyright__ = "Copyright (c) 2002-2019 {}".format(__author__)
+__copyright__ = f"Copyright (c) 2002-2019 {__author__}"
```
adiroiban, line 10: While at it :)

```suggestion
__copyright__ = f"Copyright (c) 2002-2021 {__author__}"
```

pyproject.toml
```diff
@@ -35,7 +35,7 @@ requires=
 ; * test - run trial tests
 ; * dev - run things in local dev mode
 envlist =
-    py{35,36,37,38,39,py3}-test-{twlatest,twtrunk,dev}
+    py{36,37,38,39,py3}-test-{twlatest,twtrunk,dev}
```
adiroiban, line 38: Very minor comment.
I now see it.... maybe have `py-linters` using OS default python.

Recently I started using `tox -p all` as a way to run "default" tests.

Maybe it would be nice to have it in ldapor... but I don't know how to make it work with all the twisted versions.

Maybe in envlinst have `py-test-dev, py-linters,documentation`

adiroiban (approved):
Looks good. Thanks.
Happy to see the new string formatting :)

I don't see why we need to wait for Twisted PR.

We no longer run on python3.5 here and I guess that trunk code will work just fine with 3.6 an d above.

Thanks!

pyproject.toml
graingert: will mull it over, I think it would need to be a new PR

PR #198, opened 2020-10-01
Remove repr_converter

graingert:
There's some remnants of py2 compat shims still in the codebase, one of which is repr_converter

In this PR I  remove those remnants of py2/py3 compat code.

However rather than replace repr_converter with to_unicode, I removed it.

Currently the repr_converter causes pureber etc to lie about
their content, worse when provided bytes that cannot be
decoded with utf8 (eg ObjectSID)
the reprs crash causing problems with pytest assertions

### Contributor Checklist:

### DIFF
--- a/ldaptor/ldiftree.py
+++ b/ldaptor/ldiftree.py
@@ -289,14 +289,12 @@ class LDIFTreeEntry(
 
     def __lt__(self, other):
         if not isinstance(other, LDIFTreeEntry):
-            # We don't return NotImplemented so that we get the same
-            # result in Python2 and Python3.
-            raise (TypeError, "unorderable types: {!r} > {!r}".format(self, other))
+            return NotImplemented
         return self.dn < other.dn
 
     def __gt__(self, other):
         if not isinstance(other, LDIFTreeEntry):
-            raise (TypeError, "unorderable types: {!r} < {!r}".format(self, other))
+            return NotImplemented
         return self.dn > other.dn
 
     def commit(self):
--- a/ldaptor/protocols/pureber.py
+++ b/ldaptor/protocols/pureber.py
@@ -33,7 +33,7 @@
 import sys
 from collections import UserList
 
-from ldaptor._encoder import to_bytes, repr_converter, WireStrAlias
+from ldaptor._encoder import to_bytes, WireStrAlias
 
 # xxxxxxxx
 # |/|\.../
@@ -225,7 +225,7 @@ class BEROctetString(BERBase):
         return result
 
     def __repr__(self):
-        value = repr_converter(self.value)
+        value = self.value
         if self.tag == self.__class__.tag:
             return self.__class__.__name__ + "(value=%s)" % repr(value)
         else:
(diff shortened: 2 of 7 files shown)


ldaptor/ldiftree.py
```diff
@@ -289,14 +289,12 @@ def __repr__(self):
 
     def __lt__(self, other):
         if not isinstance(other, LDIFTreeEntry):
-            # We don't return NotImplemented so that we get the same
-            # result in Python2 and Python3.
-            raise (TypeError, "unorderable types: {!r} > {!r}".format(self, other))
+            return NotImplemented
```
adiroiban, line 292: Maybe have a better error message here
```suggestion
            return NotImplemented('LDIFTreeEntry comparison operation not supported.')
```

ldaptor/ldiftree.py
```diff
@@ -289,14 +289,12 @@ def __repr__(self):
...
         return self.dn < other.dn
 
     def __gt__(self, other):
         if not isinstance(other, LDIFTreeEntry):
-            raise (TypeError, "unorderable types: {!r} < {!r}".format(self, other))
+            return NotImplemented
```
adiroiban, line 297: same as above regarding adding a more informative error message.

ldaptor/protocols/pureber.py
```diff
@@ -225,7 +225,7 @@ def toWire(self):
         return result
 
     def __repr__(self):
-        value = repr_converter(self.value)
+        value = self.value
```
adiroiban, line 228: this is ok... but maybe have a single representation for all instances

I have no idea why we need a special case here.

```
return self.__class__.__name__ + "(value=%r, tag=%d)" % (value, self.tag)
```

adiroiban (approved):
Looks good.

I am a bit worried that we have some many repr tests and we have the special handling depending on the tag value.

repr should only be used for debugging and as dev helper tools.. and we should only have the minium test coverage for repr ... and the returned value should not be part of the public API

ldaptor/ldiftree.py
graingert: `TypeError: 'NotImplementedType' object is not callable`

adiroiban:
Regarding required coverage.
I made the following coverage checks as required:

* codecov/patch - a diff should have full coverage as any new code should be merged with associated tests
* codecov/project/tests - make sure all tests are executed at least once. this should detect if we have tests which are always skipped

ldaptor/protocols/pureber.py
graingert: probably going to attrs everything next week, dunno why tag is sometimes excluded.

ldaptor/ldiftree.py
adiroiban: Sorry. Will  `NotImplementedError` do the job?

graingert: NotImplemented automatically raises a              `raise (TypeError, "unorderable types: {!r} > {!r}".format(self, other)) ` style message with nice reprs etc - it's just a bit faster

graingert: >  If all attempts return NotImplemented, the interpreter will raise an appropriate exception.

PR #204, opened 2020-10-03
remove additional use of cStringIO

graingert:
## Contributor Checklist:

looks like all these bin files need coverage - I'll move them to console_scripts shortly


adiroiban (approved):
Thanks.

It would have been nice to have a test for the script so that in the future we don't introduce regression... but this can be merged as it is.

PR #205, opened 2020-10-03
Move bin into entry points, this exposes the scripts to coverage checks

graingert:
### Contributor Checklist:


ldaptor/scripts/fetchschema.py
```diff
@@ -75,3 +73,7 @@ class MyOptions(
         baseDN=opts["base"], serviceLocationOverrides=opts["service-location"]
     )
     main(cfg)
+
+
+if __name__ == "__main__":
```
adiroiban, line 78: does it make sense to keep this here?

graingert: Yeah it's handy for `python -m ldaptor.scripts.fetchschema`

docs/source/NEWS.rst
```diff
@@ -4,7 +4,8 @@ Changelog
 20.1.2 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- fix ``ModuleNotFoundError: No module named 'cStringIO'`` in ldaptor-ldap2pdns.
+- move scripts to console_scripts entry_points
```
graingert, line 8:
```suggestion
- move scripts to console_scripts entry_points. The ``ldaptor.scripts`` package
  has been included in the standard library on a provisional basis. Backwards
  incompatible changes (up to and including removal of the module) may occur
  if deemed necessary.
```

ldaptor/scripts/fetchschema.py
adiroiban: I think that this should have a `# pragma: no cover` comment as we will not test the script via this interface.

The script can be tested via the console_script interface.

graingert: I'm currently prepping smoke tests that use this interface. Calling the function directly is difficult because of the reactor.stop calls

docs/source/NEWS.rst
graingert: or maybe it would be better in `ldaptor._scripts` as I want to change the `def main(` to be `async def main(reactor):` for `task.react`

adiroiban: well... I think that each script should have an high level entry point which will return a deferred.

and then we can have a thin wrapper which runs the deferred in reactor and stops the reactor.
This can be done via t.i.task.react https://twistedmatrix.com/documents/current/api/twisted.internet.task.html#react

graingert: Yeah I've got a PR to add corofn support to t.i.task.react

graingert: I think it's best to have it in `ldaptor._scripts` until we've nailed down a nice public interface

adiroiban, line 8: This is ok as I don't want to make ldaptor complicated... and I don't know how many people are using ldaptor.

But this release notes text looks like a commit message, rather than an information targeted to ldaptor end users.

The release notes can be something like this:

* The ldaptor scripts were converted to entry points. If you are just calling them scripts from external source nothing was changes. If you import code from the scripts, you will need to update your code as the actual scripts python code was moved into the `ldaptor._scripts` package.

adiroiban (approved):
Fine by me. You can merge it even if coverage is not 100%.

let me know if you don't have permissions and I can force merge it.

Thanks!

docs/source/NEWS.rst
graingert: I don't think any of the bin files were importable - they all contained hyphens and no `.py` extension

adiroiban: OK. I have never used the scripts directly. I tried to use them as example, but they looked complicated  :)

I have used docs/source/examples/quickstart_client.py as a good start.
I have the following script which I find it "easy" to understand: https://gist.github.com/adiroiban/d1926cfc607fd5ef588a908af61d07fd

So maybe the scripts could be redesigned to also server as documentation...
and instead of docs/source/examples/quickstart_client.py , docs/source/examples/client_basic.py and bin/ldaptor-search we can have a single reference script.

PR #206, opened 2020-10-04
use tox-gh-actions to group tox environments

graingert:
image

https://github.com/actions/toolkit/blob/1cc56db0ff126f4d65aeb83798852e02a2c180c3/docs/commands.md#group-and-ungroup-log-lines

see https://github.com/ymyzk/tox-gh-actions/issues/32
see https://github.com/tox-dev/action/issues/1
see https://github.com/tox-dev/tox-travis/issues/152
see https://github.community/t/log-group-multiplexing/130543

### Contributor Checklist:


adiroiban (approved):
All good. Thanks.

I have observed something.
Rather than having the linter skipped on 99% of the jobs, I think that is better to create a separate job, outside of the OS/python matrix which is running only on ubuntu and is calling on the linters  / documentation build steps.

PR #207, opened 2020-10-04
Enable tox --parallel 

graingert:
### Contributor Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -75,6 +75,7 @@ jobs:
       - name: tox
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TOX_PARALLEL_NO_SPINNER: 1
         run:
           tox --parallel=all
           -e'py{35,36,37,38,py3}-test-{twlatest,twtrunk},publishcov,py38-linters,release'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,7 +67,7 @@ jobs:
         if: startsWith(matrix.os, 'windows-') || startsWith(matrix.os, 'macOS-')
         run: echo '::set-env name=TOX_SKIP_ENV::pypy3.*'
 
-      - name: build coverage # coverage doesn't like being built in parallel
+      - name: build coverage for pypy3 # coverage doesn't like being built in parallel
         if: "!startsWith(matrix.os , 'windows-') || startsWith(matrix.os, 'macOS-')"
         run: |
           pypy3 -m pip install coverage>=5.3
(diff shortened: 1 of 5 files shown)


graingert:
I don't expect this to be useful without a GHA update, https://github.community/t/log-group-multiplexing/130543/5

graingert:
@adiroiban this might be nicer, what do you think? Currently we have to wait for the job to finish before seeing any output - https://github.community/t/log-group-multiplexing/130543/5 but maybe that's not so bad?

.github/workflows/main.yml
```diff
@@ -49,19 +63,27 @@ jobs:
...
           pip install tox
 
-      - name: tox linters
-        if: matrix.python-version != 3.5 && matrix.python-version != 'pypy3'
-        run: tox -e py-linters
+      - name: Disable pypy3
+        if: startsWith(matrix.os, 'windows-') || startsWith(matrix.os, 'macOS-')
+        run: echo '::set-env name=TOX_SKIP_ENV::pypy3.*'
+
+      - name: build coverage for pypy3 # coverage doesn't like being built in parallel
```
graingert, line 70:
```suggestion
      # coverage doesn't like being built in parallel, see https://git.io/JUdlT
      - name: build coverage for pypy3
```

graingert:
Takes about a minute longer overall, and the coverage change looks worrying

adiroiban (commented):
I think tox --parallel is ok for local builds.

For automated tests, I don't know if this help. I think that the VMs are 2 cores.... and a lot of tests are CPU intensive.

And is best if we see each environment as a separate status.

The coverage config change is ok.

I would say that is best to keep non-parallel runs for GHA runs.

Thanks!

graingert:
> For automated tests, I don't know if this help. 

Could do a run without `--parallel` I think mostly we save time on the pip installs

.github/workflows/main.yml
```diff
@@ -49,19 +63,27 @@ jobs:
...
+        run: |
+          pypy3 -m pip install coverage>=5.3
 
       - name: tox
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: tox -e py-test-twlatest,py-test-twtrunk,publishcov,release
+          TOX_PARALLEL_NO_SPINNER: 1
+        run:
+          tox --parallel=all
```
graingert, line 80:
```suggestion
          tox
```

graingert:
@adiroiban tox without `--parallel` is quite a bit slower

adiroiban (approved):
Thanks for the test ;)

What I wanted to say, is that for GitHub Action we can have separate jobs for each environment.

But I am ok to run one job for each OS, and inside that job to test various Python version.

You can merge it as you think is best.

Thanks!

graingert:
@adiroiban hmm the coverage has gone down only on `while True:` lines that a clearly covered... I think this might be a coverage combine bug

graingert:
@adiroiban https://github.com/nedbat/coveragepy/issues/1036

adiroiban:
Thanks. I think that this can be merged.

PR #210, opened 2020-10-06
add Python 3.9 support

graingert:
### Contributor Checklist:

- [ ] I have updated the release notes at `docs/source/NEWS.rst`
- [x] I have updated the automated tests.
- [x] All tests pass on your local dev environment. See `CONTRIBUTING.rst`.

PR #211, opened 2020-10-06
Replace use of deprecated decode encode string

graingert:
### Contributor Checklist:

- [x] I have updated the release notes at `docs/source/NEWS.rst`

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -6,6 +6,7 @@ Changelog
 
 - fix ``ModuleNotFoundError: No module named 'cStringIO'`` in ldaptor-ldap2pdns.
 - move scripts to console_scripts entry_points
+- replace deprecated calls to ``base64.decodestring`` and ``base64.encodestring``.
 
 
 20.1.1 (2020-10-02)
--- a/ldaptor/entry.py
+++ b/ldaptor/entry.py
@@ -26,7 +26,7 @@ def sshaDigest(passphrase, salt=None):
     s = sha1()
     s.update(passphrase)
     s.update(salt)
-    encoded = base64.encodestring(s.digest() + salt).rstrip()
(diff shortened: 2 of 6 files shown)


adiroiban (approved):
Looks good. Thanks.

I don't think that this change is worth being mentioned in the release notes.
We have the git log for low-level changes.

docs/source/NEWS.rst
```diff
@@ -6,6 +6,7 @@ Changelog
 
 - fix ``ModuleNotFoundError: No module named 'cStringIO'`` in ldaptor-ldap2pdns.
 - move scripts to console_scripts entry_points
+- replace deprecated calls to ``base64.decodestring`` and ``base64.encodestring``.
```
graingert, line 9: I think it's worth including the changelog - as it makes the code work on Py3.9. But I don't want to advertise py3.9 support without CI support: https://github.com/twisted/ldaptor/pull/210

```suggestion
- replace deprecated calls to ``base64.decodestring`` and ``base64.encodestring``,
  to support Python 3.9. Official support for Python 3.9 pending a new twisted release.
```

graingert:
merging this now, will decide on the changelog at the next release - eg if twisted does ship a 3.9 supporting build before we do

adiroiban:
I guess that ldaptor can have a release with support 3.9 before Twisted.

As long as the tests are green on Python 3.9 it means that ldaptor is ready for 3.9.
I don't see why ldaptor should be blocked waiting for a Twisted release.

So if #210 is green I think it can be merged and if someone wants to run ldaptor on 3.9 we can also have a new release.

docs/source/NEWS.rst
adiroiban: I would say that this PR could have been merged without a release note.

In the #210 PR the release note can read: 'Ldaptor now supports Python 3.9'

But is ok. Thanks

PR #212, opened 2020-10-08
use version attr: to extract version with declarative setuptools

graingert:
### Contributor Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,6 @@
 [metadata]
 name = ldaptor
+version = attr: ldaptor.__version__
 description = A Pure-Python Twisted library for LDAP
 long_description = file: README.rst
 author = Tommi Virtanen
--- a/setup.py
+++ b/setup.py
@@ -1,27 +1,3 @@
-#!/usr/bin/env python3
-
-import codecs
-import os
-import re
-
 from setuptools import setup
 
-
-here = os.path.abspath(os.path.dirname(__file__))
-
-
-def read(*parts):
-    with codecs.open(os.path.join(here, *parts), "r") as f:
-        return f.read()
-
-
-def find_version(*file_paths):
-    version_file = read(*file_paths)
-    version_match = re.search(r"^__version__ = ['\"]([^'\"]*)['\"]", version_file, re.M)
-    if version_match:
-        return version_match.group(1)
-    raise RuntimeError("Unable to find version string.")
-
-
-if __name__ == "__main__":
-    setup(version=find_version("ldaptor", "__init__.py"))
+setup()


graingert:
@adiroiban we can now delete the `setup.py`

adiroiban:
@graingert if `pip install -e` is no longer supported, does it mean that when you make a change and you want to run a test you need to recreate a wheel/sdist and install ?

I am always using `pip install -e .` for developmet

graingert:
Yeah I tend to use `tox py38-test-twlatest`

adiroiban:
tox is slow ...  it takes 40seconds for a second run, after all deps were already installed

In my view tox is for CI and automated test execution and not for TDD.

It takes 1 second to run all 672 tests with trial in dev mode. This is TDD friendly :)

adiroiban:
sometimes I am also using tox to setup the virtualenv with 

```sh
tox -e py38-test-dev --develop
 . .tox/py38-test-dev/bin/activate
trial ldaptor
```

graingert:
@adiroiban https://discuss.python.org/t/next-steps-for-editable-develop-proof-of-concept/4118/28

adiroiban:
@graingert ... sorry, but I don't undestand the root problem.

Why we can't have ldaptor install in dev mode and why we need extra hacks?

ldaptor has no binary module / C extension so at develop time there is no need to recompile any extension.

But, do as you think is best. My comment is just to make sure contributing to ldaptor is not a slow and painful process.

I think that `pip install -e .`  is a common tool and people are using it without having to read many extra pages of documentation and information on how to contribute to a project

graingert:
@adiroiban there's currently a movement to decouple package building from package installing:
* Previously pip would be responsible for invoking `setuptools` via `setup.py`. The goal is to move pip to installing `whl` files and using `python -m build/pep517` to create those `whl` files.
* Currently a `whl` file doesn't represent a changing/live updating directory of files - as it is just a zip file.
* So the `pip install -e .` invokes the old (pending deprecation) `setup.py` via plain setuptools without isolated builds or wheels etc.

adiroiban:
Many thanks for the detailed explanation.

my undetstanding is that `pip install -e .` is not yet deprecated and currently there is no standardized alternative.
So `python setup.py develop` via  `pip install -e .` is still the recommended way to install in dev mode.

So I think that untill there is a replacement for python setup.py develop we can keep the setup.py.
This is a small file and I don't see any issue with continue keeping it.

Thanks again for the update from this PR :)

graingert:
The `setup.py` is indeed a small file, but it exposes the legacy setuptools interface to consumers of the sdist

PR #213, opened 2020-10-14
use a pep440 direct reference to get Twisted trunk from git

graingert:
instead of legacy `#egg=`

### Contributor Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ deps =
     pyparsing
     passlib
     twlatest: Twisted[tls]
-    twtrunk: https://github.com/twisted/twisted/archive/trunk.tar.gz#egg=Twisted[tls]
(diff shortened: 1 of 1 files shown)


adiroiban (approved):
Looks good.

I think that as a drive-by, this PR can remove tw162 up to tw140 as they are no longer used.

also, `twlatest` is the same as `dev` so I think that we can remove twlatest.

And... do we need twtrunk ? I don't remember seeing ldaptor issues related to Twisted trunk in a very long time.

I think that ldaptor build can be simplified ... if we found bugs, we can add twtrunk.

thanks!

graingert:
Currently 3.9 only works in Trunk

adiroiban:
> Currently 3.9 only works in Trunk

ok. Then we will keep trunk :)

psi29a:
Nice work @graingert

PR #218, opened 2020-11-19
Fixed asText binary string handling

alinmear:
Fix LDAPFilter.asText() Problems regarding string and bytestrings:

```
builtins.TypeError: a bytes-like object is required, not 'str'
```

### DIFF
--- a/ldaptor/protocols/pureldap.py
+++ b/ldaptor/protocols/pureldap.py
@@ -33,6 +33,8 @@ def alloc_ldap_message_id():
 
 
 def escape(s):
+    if isinstance(s, (bytes, bytearray)):
+        s = s.decode()
     s = s.replace("\\", r"\5c")
     s = s.replace("*", r"\2a")
     s = s.replace("(", r"\28")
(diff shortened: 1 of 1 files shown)


graingert:
oh no, we got hit by:

```
Error: Unable to process command '::set-env name=TOX_SKIP_ENV::pypy3.*' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands
```

graingert:
let me fix that in a new pr

ldaptor/protocols/pureldap.py
```diff
@@ -33,6 +33,8 @@ def alloc_ldap_message_id():
 
 
 def escape(s):
+    if isinstance(s, (bytes, bytearray)):
+        s = s.decode()
```
graingert, line 37: > In addition, arbitrary binary data may be represented by using the escape sequence syntax by encoding each byte of binary data with the backslash () followed by two hexadecimal digits. For example, the four-byte value 0x00000004 is encoded as \00\00\00\04 in a filter string.


see the example in https://www.ietf.org/rfc/rfc2254.html#section-5

graingert:
@alinmear you'll need to run `tox -e lint`

ldaptor/protocols/pureldap.py
alinmear: thx for the hint. so we need to fix this:

```
s = s.replace("\\", r"\5c")
builtins.TypeError: a bytes-like object is required, not 'str'
```

PR #220, opened 2020-12-01
check if key in self._object_class_keys

scottcarr:
It's possible for `key` to be either `"objectClass"` or `b"objectClass"`.  Other places use this self._object_class_keys to handle this scenario.

### DIFF
--- a/ldaptor/protocols/ldap/ldapsyntax.py
+++ b/ldaptor/protocols/ldap/ldapsyntax.py
@@ -247,7 +247,7 @@ class LDAPEntryWithClient(entry.EditableLDAPEntry):
         lst = list(self.items())
         lst.sort()
         for key, values in lst:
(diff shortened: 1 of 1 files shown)


ldaptor/protocols/ldap/ldapsyntax.py
```diff
@@ -247,7 +247,7 @@ def toWire(self):
         lst = list(self.items())
         lst.sort()
         for key, values in lst:
-            if key != "objectClass":
+            if key not in self._object_class_keys:
```
graingert, line 250: just above "objectClass" is handled explicitly too, does that need to use _object_class_keys ?

```
        objectClasses = list(self.get("objectClass", []))
        objectClasses.sort()
        a.append(("objectClass", objectClasses))
```

PR #227, opened 2021-06-04
Fix folded attribute example

cl0ne:
Also clarified which attribute uses folding

### DIFF
--- a/docs/source/ldap-intro.rst
+++ b/docs/source/ldap-intro.rst
@@ -117,7 +117,7 @@ A simple LDAP file with two entries::
     telephonenumber: +1 408 555 1212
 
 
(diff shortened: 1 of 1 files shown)


docs/source/ldap-intro.rst
```diff
@@ -117,7 +117,7 @@ A simple LDAP file with two entries::
     telephonenumber: +1 408 555 1212
 
 
-A file containing an entry with a folded attribute value, from `RFC 2849 <https://www.ietf.org/rfc/rfc2849.txt>`_::
+A file containing an entry with a folded ``description`` attribute value, from `RFC 2849 <https://www.ietf.org/rfc/rfc2849.txt>`_::
```
graingert, line 120:
```suggestion
A file containing an entry with a folded ``description`` attribute value, from `RFC 2849 <https://www.ietf.org/rfc/rfc2849.html#page-8>`_::
```

PR #242, opened 2024-02-09
Use configparser.ConfigParser

s-t-e-v-e-n-k:
SafeconfigParser has been deprecated since Python 3.2, and was finally removed in 3.12, so switch to it directly.

- [x] All tests pass on your local dev environment. See `CONTRIBUTING.rst`.


psi29a:
Fixed in commit e24e29f

